PR #11, opened 2023-09-26
move the distutils shim module back to a plain module

nitzmahone:
* MANIFEST.in, the source layout, and pytest's default sys.path behavior conspired to mask a busted package, even though we were testing against the built package. `pytest` adds cwd to the path first, and since the `cffi` is in the root (grr), even though we're testing against the installed wheel, pytest loaded the subpackage from the source, masking that it was missing from the actual package (due to flat inclusions in MANIFEST.in). This is a longer-term problem that should be addressed by moving to a standard `src/` layout, and possibly also by bringing `_cffi_backend` in as a subpackage, but there are likely many dragons there with wheels that assume the presence of the top-level package.


nitzmahone:
included in #12

PR #18, opened 2023-09-29
Build wheels from sdist in CI

webknjaz:
This patch integrates `re-actors/checkout-python-sdist` in place of `actions/checkout`.
It allows making sure that the sdists are downstream-friendly by not relying on Git
checkouts for making wheels. This is how pip makes wheels when there's no matching
built distributions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,10 @@ on:
   schedule:
   - cron: '0 12 * * 1'
 
+env:
+  dists-artifact-name: artifact
+  sdist-artifact-name-wildcard: cffi-*.tar.gz
+
 jobs:
   sdist:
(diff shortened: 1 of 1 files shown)


nitzmahone:
Implemented by #79

webknjaz:
@nitzmahone partially, right?

nitzmahone:
@webknjaz All the wheels are being built from the same-run sdist artifact now, unless I missed something? The only checkouts now are the sdist job itself and the matrix generator jobs, since those need access to the (currently) in-repo custom action. Once that's settled down, I might externalize it (esp since I'm using it in PyYAML too), but want to see how it holds up in the real world for awhile.

webknjaz:
@nitzmahone this is what I had in mind: https://github.com/python-cffi/cffi/pull/82.

PR #19, opened 2023-09-29
draft implement dynamic matrix

nitzmahone:


ofek:
It looks like this needs to be rebased.

nitzmahone:
Implemented "for real" by #79

PR #22, opened 2023-10-13
Fix dead setuptools links

JamesTheAwesomeDude:
As of Python 3.12, the setuptools documentation is maintained out-of-tree and has been removed from docs.python.org.

### DIFF
--- a/doc/source/cdef.rst
+++ b/doc/source/cdef.rst
@@ -80,7 +80,8 @@ In order of complexity:
 
 * Finally, you can (but don't have to) use CFFI's **Distutils** or
   **Setuptools integration** when writing a ``setup.py``.  For
-  Distutils (only in out-of-line API mode):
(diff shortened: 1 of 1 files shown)


JamesTheAwesomeDude:
Additional context: https://peps.python.org/pep-0632/

doc/source/cdef.rst
```diff
@@ -94,7 +95,7 @@ In order of complexity:
         ext_modules=[foo_build.ffibuilder.distutils_extension()],
     )
 
-  For Setuptools (out-of-line, but works in ABI or API mode;
+  For Setuptools (out-of-line only, but works in ABI or API mode;
```
JamesTheAwesomeDude, line 98: (Correct me if I'm wrong here.)

arigo:
Thanks for the link fixes and the clarification!

PR #30, opened 2023-11-03
parallel build support for ffibuilder.compile()

JamesTheAwesomeDude:
Is there any real reason this wasn't added yet?

```python
from cffi import FFI

from multiprocessing import cpu_count

ffibuilder = FFI()
...
ffibuilder.compile(verbose=True, parallel=multiprocessing.cpu_count())
```

### DIFF
--- a/src/cffi/api.py
+++ b/src/cffi/api.py
@@ -707,7 +707,7 @@ class FFI(object):
         recompile(self, module_name, source,
                   c_file=filename, call_c_compiler=False, **kwds)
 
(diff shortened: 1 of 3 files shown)


arigo:
Looks reasonable to me.  It's missing tests and documentation for all public methods with the new `parallel` keyword argument.  (For tests I'd be happy with just making sure there is at least one test somewhere that gives some value to `parallel`, and one that doesn't, and both passes.)

PR #34, opened 2023-11-15
Use PyErr_FormatUnraisable() on Python 3.13

vstinner:
Use the new public PyErr_FormatUnraisable() on Python 3.13.

The private _PyErr_WriteUnraisableMsg() function was removed in Python 3.13:
https://github.com/python/cpython/pull/111643

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -6118,7 +6118,11 @@ static void _my_PyErr_WriteUnraisable(PyObject *t, PyObject *v, PyObject *tb,
 
     PyErr_Restore(t, v, tb);
     if (s != NULL) {
+#if PY_VERSION_HEX >= 0x030D0000
+        PyErr_FormatUnraisable("Exception ignored %S", s);
+#else
         _PyErr_WriteUnraisableMsg(PyText_AS_UTF8(s), NULL);
+#endif
         Py_DECREF(s);
     }
     else


src/c/_cffi_backend.c
```diff
@@ -6109,6 +6108,18 @@ static void _my_PyErr_WriteUnraisable(PyObject *t, PyObject *v, PyObject *tb,
     if (extra_error_line == NULL)
         extra_error_line = "";
 
+#if PY_VERSION_HEX >= 0x030D0000
+    if (obj != NULL) {
+        PyErr_FormatUnraisable("Exception ignored %c%s%R%s",
```
vstinner, line 6113: Maybe it's an opportunity to enhance the error message?

Should it be "Exception ignored ...", "Exception ignored **in** ..." (add `in `) or  "Exception ignored: ..." (add `: `)?

vstinner:
cc @serhiy-storchaka @arigo

vstinner:
By the way, `_PyThreadState_UncheckedGet()` should be replaced with the new public `PyThreadState_GetUnchecked()` function.

serhiy-storchaka (commented):
You forget to add `PyErr_Restore()`.

src/c/_cffi_backend.c
```diff
@@ -6109,6 +6108,18 @@ static void _my_PyErr_WriteUnraisable(PyObject *t, PyObject *v, PyObject *tb,
...
+                               first_char, objdescr + 1, obj,
+                               extra_error_line);
+    }
+    else {
+        PyErr_FormatUnraisable("Exception ignored %c%s%s",
+                               first_char, objdescr + 1, extra_error_line);
+    }
+#else
+    PyObject *s;
```
arigo, line 6122: This does not compile on older Visual Studio versions, but I'm unsure how old these need to be.  You can't declare a variable anywhere else than at the start of a block in older versions of C.

vstinner:
> You forget to add PyErr_Restore().

Oops, fixed.

src/c/_cffi_backend.c
vstinner: Ok, fixed. I made the assumption that cffi is built by a C99 compiler these days.

arigo:
@serhiy-storchaka has a point.  And I'd be happier with some change that avoids making two versions of the formatting code.  As far as I'm concerned, killing a semi-private API is a minor annoyance that I'd be more happy to work around, instead of trying to rewrite more CFFI logic, because that ends up making duplication.  So I wouldn't kill `PyUnicode_FromFormat()` just because we appear to be able to in some CPython version.

src/c/_cffi_backend.c
arigo: I'm trying to keep the code compiling for 2.7, and that might require much older VS on Windows.

vstinner:
@arigo:
> @serhiy-storchaka has a point. And I'd be happier with some change that avoids making two versions of the formatting code. As far as I'm concerned, killing a semi-private API is a minor annoyance that I'd be more happy to work around, instead of trying to rewrite more CFFI logic, because that ends up making duplication. So I wouldn't kill PyUnicode_FromFormat() just because we appear to be able to in some CPython version.

Sure, it makes sense. I rewrote my PR to also use the *s* string on Python 3.13. The patch is now just 4 lines.

@arigo: Would you mind to review the updated PR?

src/c/_cffi_backend.c
vstinner: I'm trying hard for provide a best effort support for Python 2.7 in my https://pythoncapi-compat.readthedocs.io/ project. I had C code to support VS2015 which doesn't support C99 `static inline`. But I removed it, it made the code harder to maintain. Moreover, GitHub Action doesn't provide Python 2.7. I'm now considering to officially remove Python 2 support and remove Python 2 code in pythoncapi-compat.

nitzmahone: Yep, I've tried "best effort" support like this so many times, but even with the best of intentions, it's effectively always broken when not backed by automated testing. It also ends up actively discouraging contributions (like this one!)...

arigo: @nitzmahone Sure, I see the point, but in this case the final result is that the pull request was changed to 4 obvious lines instead of a more complex #if dance that missed originally a `PyErr_Restore()`, in addition to breaking a compiler (that may or may not be too old to consider).  I think it's a win for a project like CFFI.

arigo:
Sure.  Now I have no hesitation to merge the pull request.  @nitzmahone, can you please confirm if I should "Enable auto-merge (squash)" for this kind of case?

nitzmahone:
> can you please confirm if I should "Enable auto-merge (squash)" for this kind of case?

Yeah, that's fine- basically "I'm happy with the code, merge on my behalf if all checks pass" vs babysitting it...

vstinner:
Thanks for the review and merging my change! I first wanted to *enhance* the error message, but I understand the desire to have the same message on all Python versions and it makes sense.

Ah yeah, I missed `PyErr_Restore()`, since I'm used to call `PyErr_WriteUnraisable()` while an exception in set, not on a function which gets an exception as an argument. Hopefully, it was catched by review and fixed ;-)

nitzmahone:
@vstinner @arigo Bleh- I threw together a manylinux_2_28 container with 3.13.0a1 in it and hacked up `cibuildwheel` to work with it (in hopes of wiring up a 3.13 prerelease canary build here), and it's failing under 3.13 with an unresolved symbol for `PyErr_FormatUnraisable`...

Shouldn't this have been `PyErr_WriteUnraisable`, not `PyErr_FormatUnraisable`? I don't see any libpython exports for FormatUnraisable:
```
(tmp-ad2272c74e493a7) [mdavis@mdavis-p1 lib]$ objdump -t libpython3.13.so | grep 'PyErr_.*Unrais'
00000000002237e0 l     F .text  00000000000001b8              _PyErr_WriteUnraisableDefaultHook
00000000002239a0 g     F .text  00000000000007aa              _PyErr_WriteUnraisableMsg
0000000000224150 g     F .text  000000000000000a              PyErr_WriteUnraisable
```

nitzmahone:
(or maybe the new libpython export didn't make it into 3.13.0a1 yet)- in any case, this doesn't appear to be fully working yet...

nitzmahone:
OK, looks like this will be fixed in 3.13.0a2 (I just built against the top of `main`, which I assume will be a2 tomorrow-ish):
```
(tmp-2fcf66c976036a9) [mdavis@mdavis-p1 tmp-2fcf66c976036a9]$ objdump -t ~/.pyenv/versions/3.13-dev/lib/libpython3.13.so | grep PyErr_.*Unrai
0000000000245d00 l     F .text  00000000000001b8              _PyErr_WriteUnraisableDefaultHook
0000000000245ec0 g     F .text  0000000000000096              PyErr_FormatUnraisable
0000000000245f60 g     F .text  0000000000000009              PyErr_WriteUnraisable
```

nitzmahone:
... and yep, full test suite passes against that build of "almost-alpha2", so I think we're good to go here...
```============ 1951 passed, 89 skipped, 4 xfailed, 10 warnings in 255.53s (0:04:15) ============```

PR #52, opened 2024-01-27
use public Py_HashPointer API on 3.13+

### LINKED ISSUE #51: New breakage from 3.13.0a3
Looks like there's at least one more C API change introduced by 3.13.0a3 we'll need to account for- it builds OK, but there's a new missing private symbol at import-time:

```
ImportError: /home/mdavis/.virtualenvs/tmp-65f3b135fc600df/lib/python3.13/site-packages/_cffi_backend.cpython-313-x86_64-linux-gnu.so: undefined symbol: _Py_HashPointer
```

nitzmahone:
fixes #51

* use newly-public `Py_HashPointer` API on 3.13+

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -2457,7 +2457,11 @@ static Py_hash_t cdata_hash(PyObject *v)
         }
         Py_DECREF(vv);
     }
+#if PY_VERSION_HEX >= 0x030D0000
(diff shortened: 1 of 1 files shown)


nitzmahone:
:facepalm: on writing the conditional backwards- this is why I shouldn't do things on Friday afternoons. Thanks for the fix @arigo ! Now that CI and the private 3.13 support images/tools are up-to-date and confirm it's working properly under 3.13.0a3, merging...

PR #53, opened 2024-01-29
Use libffi in emscripten

hoodmane:
Emscripten has a libffi port and `malloc_closure.h` won't work without significant extra work in Emscripten so use the libffi implementation.


arigo:
Thanks for the fix!  Unless I'm mistaken, Emscripten doesn't support fork() anyway and has got no way of supporting it in the future.  That makes the reasoning for using the custom `malloc_closure.h` not apply (in the comment a few lines above your changes).

hoodmane:
Thanks for the quick merge @arigo!

>  Unless I'm mistaken, Emscripten doesn't support fork() anyway

Actually, it _does_ support threads:
https://emscripten.org/docs/porting/pthreads.html
I don't think the issues apply here anyways, and I always do single threaded builds so I guess someone else can investigate if they are concerned...

arigo:
Yes, threads are fine.  The issue with libffi on Linux comes from calling fork(), which is its own different beast.

Issue #56, opened 2024-02-06
Complex number support using msvc


chris-hld:
In API mode, compiling on github actions, using `_Complex` works perfectly for Linux and macOS latest.
However, on windows-latest, I am not sure how to use the interface type complex. 
I am aware that msvc doesn't support `float _Complex`, but defines its own `_Fcomplex`.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support?view=msvc-170
This is, however, not supported by cffi as it seems, since trying 
```
if sys.platform != "win32":
    ffibuilder.cdef("""
                    
    typedef float _Complex float_complex;
    typedef double _Complex double_complex;

    """)
else:
    ffibuilder.cdef("""
                    
    typedef _Fcomplex float_complex;
    typedef _Dcomplex double_complex;

    """)
```
fails with 
```
 cffi.CDefError: cannot parse "typedef _Fcomplex float_complex;"
<cdef source string>:3:23: before: float_complex
Error: Process completed with exit code 1.
```

If I don't use `_Fcomplex`, only using `float _Complex` throughout the project, then msvc fails with the typical
```
build\temp.win-amd64-cpython-310\Release\_safpy.c(727): error C2143: syntax error: missing ')' before '*'
build\temp.win-amd64-cpython-310\Release\_safpy.c(727): error C2143: syntax error: missing '{' before '*'
build\temp.win-amd64-cpython-310\Release\_safpy.c(727): error C2059: syntax error: 'type'
build\temp.win-amd64-cpython-310\Release\_safpy.c(727): error C2059: syntax error: ')'
build\temp.win-amd64-cpython-310\Release\_safpy.c(736): error C2143: syntax error: missing ';' before '*'
build\temp.win-amd64-cpython-310\Release\_safpy.c(761): error C2146: syntax error: missing ')' before identifier '_Complex'
build\temp.win-amd64-cpython-310\Release\_safpy.c(761): error C2059: syntax error: ')'
build\temp.win-amd64-cpython-310\Release\_safpy.c(761): error C2100: illegal indirection
build\temp.win-amd64-cpython-310\Release\_safpy.c(761): error C2100: illegal indirection
build\temp.win-amd64-cpython-310\Release\_safpy.c(761): error C2297: '*': not valid as right operand has type 'void *'
build\temp.win-amd64-cpython-310\Release\_safpy.c(782): warning C4133: 'function': incompatible types - from 'int ***' to 'float_complex ***'
[...]

PR #57, opened 2024-02-07
Support for complex types with MSVC, called _Fcomplex and _Dcomplex

arigo:


chris-hld:
Fixes https://github.com/python-cffi/cffi/issues/56 for [my setup.](https://github.com/chris-hld/SAFpy/blob/master/safpy_build.py)

mattip:
Should this be adapted as a PR to the PyPy backend as well?

arigo:
@mattip Good point!  I think a minimal change would be to copy the 2 lines added to each of `commontypes.c` and `parse_c_type.c`.  PyPy uses these C files directly in `pypy/module/_cffi_backend/src/`.  I think that's all that is required.  Some of the tests can also be ported, though.  I can give it a try (which mostly means trying to use github to make a PR to a large project like PyPy over a not-always-fast Internet connection).

mattip:
> I can give it a try

Thanks. I opened pypy/pypy#4890 to track the problem and marked it for the next release.

PR #24, opened 2023-10-17
[WIP] Python 3.13 compatibility

nitzmahone:
fixes #23

* updates obsolete private API aliases removed in 3.13
* make unraisable tests less brittle to traceback formatting changes across Python versions

Full test suite passes locally on 3.13.0a1 under Linux x86_64 with these changes.

PR #59, opened 2024-02-15
rearrange code to make PyPy testing happier

mattip:
PyPy testing has a check that the file `src/c/test_c.py` is properly synced to the PyPy version. It checks only code below the `# __________________` marker. The `_assert_unraisable` function added in #24 needs a bit of work to support all the python versions supported by PyPy (specifically, PyPy still uses python2.7) so it should appear above the marker.

The change should be harmless in this repo since it is "only" a rearrangement of code.

cc @arigo

### DIFF
--- a/src/c/test_c.py
+++ b/src/c/test_c.py
@@ -30,6 +30,36 @@ except ImportError:
         pytest.skip("_testunc() not available")
 from _cffi_backend import __version__
 
+
+@contextlib.contextmanager
+def _assert_unraisable(error_type: type[Exception] | None, message: str = '', traceback_tokens: list[str] | None = None):
+    """Assert that a given sys.unraisablehook interaction occurred (or did not occur, if error_type is None) while this context was active"""
+    raised_errors: list[Exception] = []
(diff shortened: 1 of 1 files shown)


arigo:
A bit confused: I thought that the check you mention is disabled by now.  If it isn't, then is this commit enough?  I had to do a couple more tweaks, like disabling the tests that end up calling this function...  You are you talking about `pypy/module/_cffi_backend/test/_backend_tests.py`, right?  Not the other copy that ends up in `extra_tests`?

mattip:
Yes, this is for the untranslated PyPy `_cffi_backend` tests. The [test_file check](https://buildbot.pypy.org/summary/longrepr?testname=test_same_file&builder=own-linux-x86-64&build=10065&mod=pypy.module._cffi_backend.test.test_file) is still active (and failing) on PyPy.

arigo:
I'd consider skipping that test instead...

mattip:
I could skip it. But on the other hand, this is a pretty simple change to allow the test to keep running.

PR #69, opened 2024-02-27
Build `aarch64` `musllinux` wheel

### LINKED ISSUE #41: Add `aarch64` `musllinux` wheel
When one runs Alpine Linux (e.g., a container) on a modern MacBook, there's no binary wheel available for `cffi`. It can cause build failures in environment set up to avoid compilations from source/sdists. Can you add a wheel for this setup?

harryzcy:
[super-linter](https://github.com/super-linter/super-linter) is preparing for arm64 support https://github.com/super-linter/super-linter/issues/5070. It would be very helpful to have a `musllinux_aarch64` wheel

Fix #41

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,7 +57,6 @@ jobs:
           # HACK: private musllinux_1_1 container build with 3.13 prerelease support
           manylinux_img: quay.io/rolpdog/musllinux_1_1_x86_64:latest
 
-
         - spec: cp39-musllinux_i686
         - spec: cp310-musllinux_i686
         - spec: cp311-musllinux_i686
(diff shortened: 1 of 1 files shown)


harryzcy:
@arigo @nitzmahone Can you help to review this? Thank you

arigo:
I have no idea how `.github/workflows/ci.yaml` works, but if Matt cannot do the review for the time being, can you

harryzcy:
> it looks like it's only adding support to run on some extra CI system of github, right?

Yes

> Is there any side-effect to take into account for us?

I'm not aware of any. I don't think it effects existing wheels for other os/architecture, since the build happens separately for each system.

Thanks

harryzcy:
@arigo Thanks!

arigo:
:-)

nitzmahone:
Sorry, been buried the past few weeks leading up to a vacation (first day back)...

I should've commented earlier: I'd been trying to avoid adding new support for wheel builds on arches where we don't have native builders. The full test suite is prohibitively slow on emulated builds (and Windows, for reasons I still haven't had time to investigate), so these wheels will be effectively untested beyond the very tiny slice of the test suite that we run on the Windows wheels and anything non-Intel or Apple Silicon. Caveat Emptor :wink:

nitzmahone:
PS, been working a bit with @samdoran on fixing up the test fixture isolation to work properly with `pytest-xdist` - that should allow us to run the tests much faster in parallel worker processes.

harryzcy:
Unrelated to this PR, `macos-14` would provide Apple Silicon and it has no intel-mac as a standard runner. I don't see how other linux non-intel arches would be possible without emulation.

PR #78, opened 2024-05-23
avoid null-pointer-subtraction error

nitzmahone:
Newer compilers generate warnings/errors about null pointer subtractions being an undefined behavior. Use matching non-zero offsets (as suggested by arigo) to avoid the error.


nekopsykose:
is there any reason to not make this `offsetof(%s, %s)` instead?

arigo:
We can't in this case, because we have the name of the *pointer* type, and not the *struct* name type.  See the next line after the `else:` for the case where we have the struct name and can just emit `offsetof`.

nekopsykose:
that makes sense, thanks :)

PR #81, opened 2024-05-28
add yet another flag to recompile() to avoid calling ffiplatform

nitzmahone:


inklesspen:
Is there anything I can do to help this get merged?

arigo:
Trying to close and re-open this PR to force the checks to re-execute

arigo:
Nope, the musllinux tests still fail here.  I'm going to assume it's unrelated to the current PR and merge anyway.

nitzmahone:
Yeah, at a glance it looks like maybe the latest round of tool upgrades in the musllinux containers has broken the static libffi linkage again, and `cibuildwheel` recently started forcing musllinux 1.2.

On the plus side, I'm *so* glad @arigo came up with an easy way to alias the ffi_call stuff so an accidental non-static extension build is clearly broken instead of silently dynamic-linking in the build env, passing tests, then failing in weird ways in the real world. :tada: 

IIRC Alpine has a hacky script approximation of ldconfig that doesn't cover all the bases, and getting the setuptools extension builds to uniformly respect pkg-config is ... tricky. Opened #114 to track that issue separately.

PR #82, opened 2024-05-28
🧪 Pass sdist into cibuildwheel directly

webknjaz:
This patch also brings a copy of `cibuildwheel` in through the GitHub
Actions interface instead of PyPI.


webknjaz:
@nitzmahone while working on this I also identified invalid output references in the artifact download steps. Will try to address that now.

webknjaz:
@nitzmahone I think https://github.com/python-cffi/cffi/commit/2bea383f2cdbcf805262c2561282aaf58b53a263 constitutes a release blocker.

.github/workflows/ci.yaml
```diff
@@ -258,14 +260,14 @@ jobs:
       id: fetch_sdist
       uses: actions/download-artifact@v4
       with:
-        name: ${{ needs.build_sdist.outputs.artifact_name }}
```
webknjaz, line 261: *(this was an empty string because there's no job with ID `build_sdist`; causing all the artifacts to be downloaded and unpacked in subdirectories)*

.github/workflows/ci.yaml
```diff
@@ -404,30 +403,29 @@ jobs:
...
         CIBW_TEST_REQUIRES: pytest setuptools
         CIBW_TEST_COMMAND: pip install pip --upgrade; cd {project}; PYTHONUNBUFFERED=1 pytest
         MACOSX_DEPLOYMENT_TARGET: ${{ matrix.deployment_target || '10.9' }}
         SDKROOT: ${{ matrix.sdkroot || 'macosx' }}
-      run: |
-        set -eux
-        
-        mkdir cffi
-        
-        tar zxf ${{ steps.fetch_sdist.outputs.download-path }}/cffi*.tar.gz/cffi*.tar.gz --strip-components=1 -C cffi
```
webknjaz, line 420: *(with the requested GHA artifact name being non-empty, the new path is actually `${{ steps.fetch_sdist.outputs.download-path }}/cffi*.tar.gz`)*

.github/workflows/ci.yaml
```diff
@@ -404,30 +403,29 @@ jobs:
...
+      with:
+        package-dir: >-
```
webknjaz, line 415: *(`cibuildwheel` unarchives sdist tarballs if supplied in place of a directory)*

.github/workflows/ci.yaml
```diff
@@ -291,24 +293,21 @@ jobs:
         CIBW_PRERELEASE_PYTHONS: 'True'
         CIBW_TEST_REQUIRES: pytest setuptools  # 3.12+ no longer includes distutils, just always ensure setuptools is present
         CIBW_TEST_COMMAND: PYTHONUNBUFFERED=1 python -m pytest ${{ matrix.test_args || '{project}' }}  # default to test all
-      run: |
-        set -eux
-        
-        mkdir cffi
-        
-        tar zxf ${{ steps.fetch_sdist.outputs.download-path }}/cffi*.tar.gz/cffi*.tar.gz --strip-components=1 -C cffi
-        python -m pip install --upgrade "${{ matrix.cibw_version || 'cibuildwheel' }}"
```
webknjaz, line 300: *(I didn't keep any `${{ matrix.cibw_version }}` because it's likely some legacy leftover that doesn't actually exist anywhere)*

.github/workflows/ci.yaml
```diff
@@ -291,24 +293,21 @@ jobs:
...
+      run: echo "artifact_name=$(ls ./wheelhouse/)" >> "${GITHUB_OUTPUT}"
+      shell: bash -eEuxo pipefail {0}
 
     - name: upload artifacts
       uses: actions/upload-artifact@v4
       with:
-        name: ${{ steps.build.outputs.artifact_name }}
-        path: dist/*.whl
+        name: ${{ steps.built-artifact-lookup.outputs.artifact_name }}
+        path: wheelhouse/${{ steps.built-artifact-lookup.outputs.artifact_name }}
```
webknjaz, line 310: *(I replaced a wildcard with the known value, since we actually have it in a var already)*

.github/workflows/ci.yaml
```diff
@@ -29,6 +29,8 @@ env:
 jobs:
   python_sdist:
     runs-on: ubuntu-22.04
+    outputs:
+      artifact_name: ${{ steps.build_sdist.outputs.artifact_name }}
```
webknjaz, line 33: *(this was another missing bit that prevented the wheel jobs from referencing the tarball name)*

.github/workflows/ci.yaml
nitzmahone: Ugh- another bug introduced by moving back and forth between PyYAML and CFFI trying to standardize the workflows between them. GHA doesn't make it easy to debug these kinds of things because so many different layers just silently succeed with glaring issues like this. Their (lack of) expression type system also makes me appreciate Jinja *so* much more.

.github/workflows/ci.yaml
nitzmahone: Yeah- the artifact upload/download pathing is painful. It's even worse on Windows, since the GHA context/event properties are *usually* (but not always) Windows paths with drive letters and backslashes, and there aren't enough expression conversion operators to (reliably) convert those. Using bash on Windows for "glue" instead of cmd/PS helps some, but there were still some things that just plain didn't work properly, which is why I had to fall back to not using the artifact fetch paths in all cases.

.github/workflows/ci.yaml
nitzmahone: I'm likely going to move the libffi build to its own cached-artifact job though, which means we'll still have to give cibuildwheel a full directory to get the cached artifact dir into the container envs alongside the sdist- that's why I abandoned just giving cibuildwheel the sdist tarball directly. I looked over the code- it didn't look like there was a way to build directly from an sdist tarball *and* send a project directory with build deps at the same time.

.github/workflows/ci.yaml
nitzmahone: We should leave support in for that in some form- they're necessary every time cibuildwheel doesn't support something we need to (eg, every pre-beta Python, or when cibuildwheel drops support for a target Python before we do). I don't want to have to plumb it back in every time, since we're always testing new Pythons well before cibuildwheel has released support for them, and I haven't gotten around to PRing any of my "bring your own Python" hacks to cibuildwheel.

webknjaz: Can I do that in a follow-up? There's no structure for it left in the matrix even. Pretty sure I could sketch a better solution anyway.

.github/workflows/ci.yaml
webknjaz: Any reason those things shouldn't just be in the sdist? It's supposed to be self-contained.

.github/workflows/ci.yaml
webknjaz: Pretty sure bash can handle those. I had to work around this problem in some action already.

.github/workflows/ci.yaml
webknjaz: Interestingly, in the past, their YAML parser would error out if you were to reference an identifier not listed in the `needs:` key. It used to be rather hard to find that web page that displays the error too since the checks wouldn't appear in the PR.

webknjaz:
@mattip I don't remember but I should probably rebase if you're ready to look into it.

PR #86, opened 2024-06-09
Avoid Py_FileSystemDefaultEncoding in 3.12+

rtobar:
This has been [deprecated](https://docs.python.org/3/whatsnew/3.12.html#id6), and hence emits build warnings, which will eventually become errors.

Thus avoid it using the same mechanism that's already being used in similar situations.

First seen in https://github.com/ICRAR/ijson/actions/runs/8996107055/job/24712097518#step:7:81

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -204,6 +204,10 @@
 # define USE_WRITEUNRAISABLEMSG
 #endif
 
+#if PY_VERSION_HEX >= 0x030c0000
+# define Py_FileSystemDefaultEncoding PyConfig.filesystem_encoding
+#endif
+
 /************************************************************/
 
 /* base type flag: exactly one of the following: */


rtobar:
This is a gentle ping to see if this can be considered for merging

arigo:
It fails to build, though, at least in the automated tests on CPython 3.13 on Linux...

arigo:
`PyConfig` is a type name, not a global structure name.  I couldn't find much information about how we're supposed to replace `Py_FileSystemDefaultEncoding`, but it seems to be more involved than this PR does.

rtobar:
@arigo ouch, sorry about that! From the very simple deprecation notes I assumed that `PyConfig` was a global object that one could inspect.

Like you, doing some digging left me a bit confused as to how to do this properly. One option seems to be to invoke `sys.getfilesystemencoding()`, which seems a tad too complex, but is what one would do from Python. The other looks like declaring a `PyConfig config` objet on the stack, call `PyConfig_Read(&config)`, and then inspect the correct attribute. The documentation of the function isn't crystal clear to me though, and it would seem like the whole.mechanism is more expensive that what it should: the interpreter already has a `PyConfig` object initialised (but there's no way to access it from the C API), which is what `sys` uses internally, so we shouldn't need to create and initialise/read one again.

I'd be happy to do something in the direction of invoking `sys`, does that sound ok?

arigo:
That sounds over-complicated.  I'd rather go with a few extra lines that are at least explicit (untested code, assuming Python 3.x):

``` C
    if (!PyArg_ParseTuple(args, "U|i:load_library", &s, &flags))
        return NULL;
    *p_temp = PyUnicode_EncodeFSDefault(s);   // there's already logic to free this later
    if (*p_temp == NULL)
        return NULL;
    filename_or_null = PyBytes_AS_STRING(*p_temp);
```

rtobar:
Mmmm.m.. I reworked this quite a bit , but was waiting for the pre-3.8 C code removal before putting an update in this PR. I of course forgot that it automatically updates if I push to my repo branch, which I did to double-check that win/macos builds were working (I tested all individual commits locally in Linux).

I'll mark this PR as Draft to signal that I'm not looking to merge this until the pre-3.8 C code removal is merged; after that I'm happy to discuss this in further detail.

PR #99, opened 2024-07-25
Fix undefined names in Python code

cclauss:
% `ruff check --select=E9,F63,F7,F82 --output-format=concise`

And then fix the resulting undefined names except in file `testing/cffi1/test_re_python.py`.


arigo:
I'm fine with fixes to the old demos, but I'd prefer not to have random comments show up in the source code of cffi itself.

cclauss:
OK.  Only in `demo` files.

PR #102, opened 2024-07-26
Add GitHub Action to run codespell and ruff

cclauss:
Put automated testing in place to cover:
* #97
* #98
* #99
* #100


arigo:
No idea if this integrates with what we already run or not.  @nitzmahone, feel free to close this if it is pointless from your point of view.

PR #103, opened 2024-07-27
ruff check --select=E --fix --unsafe-fixes

cclauss:
Let's consider this ___after___ #102 is merged so we can ensure that the pytests continue to pass with these changes.
* #102

Which [ruff rules E](https://docs.astral.sh/ruff/rules/#error-e) can be automatically fixed?

% `ruff check --statistics | grep "\[\*\]"`
```
  58	E401	[*] multiple-imports-on-one-line
  51	F841	[*] unused-variable
  25	E711	[*] none-comparison
   8	E731	[*] lambda-assignment
   5	E703	[*] useless-semicolon
   2	E712	[*] true-false-comparison
   2	E713	[*] not-in-test
```
% `ruff check --select=E --fix --unsafe-fixes`
```
Found 276 errors (100 fixed, 176 remaining).
```
% `ruff rule E731 ` # https://docs.astral.sh/ruff/rules/lambda-assignment
# lambda-assignment (E731)

Derived from the **pycodestyle** linter.

Fix is sometimes available.

## What it does
Checks for lambda expressions which are assigned to a variable.

## Why is this bad?
Per PEP 8, you should "Always use a def statement instead of an assignment
statement that binds a lambda expression directly to an identifier."

Using a `def` statement leads to better tracebacks, and the assignment
itself negates the primary benefit of using a `lambda` expression (i.e.,
that it can be embedded inside another expression).

## Example
```python
f = lambda x: 2 * x
```

Use instead:
```python
def f(x):
    return 2 * x
```

[PEP 8]: https://peps.python.org/pep-0008/#programming-recommendations

### DIFF
--- a/demo/cffi-cocoa.py
+++ b/demo/cffi-cocoa.py
@@ -60,14 +60,14 @@ NSApplicationActivationPolicyRegular = ffi.cast('NSApplicationActivationPolicy',
 NSTitledWindowMask = ffi.cast('NSUInteger', 1)
 NSBackingStoreBuffered = ffi.cast('NSBackingStoreType', 2)
 
-NSMakePoint = lambda x, y: ffi.new('NSPoint *', (x, y))[0]
-NSMakeRect = lambda x, y, w, h: ffi.new('NSRect *', ((x, y), (w, h)))[0]
+def NSMakePoint(x, y):
+    return ffi.new('NSPoint *', (x, y))[0]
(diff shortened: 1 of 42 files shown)


arigo:
I know I didn't follow the letter of PEP 8 when writing CFFI.  It's far too late now.  I'm not reviewing changes to 42 files just for stylistic purposes, sorry.  Additionally, CFFI is mostly in maintenance mode only.

Issue #40, opened 2023-12-07
Support for new `cp313t` GIL-free ABI variant


nitzmahone:
I did some smoke testing with the new GIL-free `t` CPython ABI on 3.13.0a2. I had to patch up `wheel` with a pre-release version of `packaging` to get it to build and install without barfing on the new ABI tag, but once I got past that, *most* of the CFFI test suite is passing. Looks like the majority of the failures are related to new symbol exports (possibly unique to the nogil build?) that need to be accounted for, eg:
```
ERROR testing/cffi1/test_new_ffi_1.py::TestNewFFI1::test_addressof_anonymous_struct - 
ImportError: /tmp/ffi-4/test_new_ffi_1.cpython-313t-x86_64-linux-gnu.so: 
undefined symbol: _Py_atomic_load_uint32_relaxed
```

I've already added some 3.13 pre-release smoke-testing- once the various packaging tools we need have released versions that support the new `t` ABI tagging, we can include that as well.

arigo:
It's very much more involved than that.  CFFI is full of static global caches (guaranteeing the uniqueness of many CTypeObject, for example).  There is also careful logic around when to acquire and release the GIL around calls, and particularly callbacks.  I believe that if you just try out the tests, then yes, it will mostly work because they are mostly single-threaded.  In a real multithreaded environment you're going to see very obscure failures sometimes (e.g. "object of ctype 'int **' is expected, but got instead 'int **'").

Supporting multithreading requires a through review of the whole source code of CFFI, I'm afraid.  Otherwise we get some "meh, doesn't usually crashes, that's good enough".  I certainly hope that it's not what CPython is going for in the GIL-free ABI variant.  I would object to distribute such a version of CFFI.

nitzmahone:
@arigo 100% agreed on all points- I just wanted to see what kind of gaps might need to be filled to even get it to build and run single-threaded under the new `t` ABI.

I re-tried some of the concurrency testing I'd previously done on Sam's 3.9 `nogil` PoC branch against both 3.13.0a2 and 3.13 head just to see how unstable things were, but the GIL disablement doesn't even appear functional with a `--disable-GIL` interpreter build yet. IIUC, that part [hasn't even merged yet](https://github.com/python/cpython/issues/108219) (see the "Enable/Disable the GIL at runtime" commit), so none of the real investigation or testing can even start beyond "yay, it builds".

[...]

PR #104, opened 2024-08-01
limited api is not available in free-threaded cpython

minrk:
related to #40

Free-threaded builds [don't support py_limited_api](https://docs.python.org/3.13/howto/free-threading-extensions.html#limited-c-api-and-stable-abi). Even setting `py_limited_api=False` still defines py_limited_api unless `_CFFI_NO_LIMITED_API` is _also_ defined (fixed here), so compilation would still fail with:

```
$PREFIX/include/python3.13t/Python.h:51:4: error: "The limited API is not currently supported in the free-threaded build"
#  error "The limited API is not currently supported in the free-threaded build"
```

Perhaps it is also bug that `set_source(py_limited_api=False)` still creates a source that defines Py_LIMITED_API on most systems?

(this is by no means "free-threaded support" for cffi, but one small step required to build cffi extensions _at all_ on free-threaded cpython)

### DIFF
--- a/src/cffi/recompiler.py
+++ b/src/cffi/recompiler.py
@@ -1,4 +1,5 @@
 import os, sys, io
+import sysconfig
 from . import ffiplatform, model
 from .error import VerificationError
 from .cffi_opcode import *
@@ -8,7 +9,8 @@ VERSION_EMBEDDED = 0x2701
 VERSION_CHAR16CHAR32 = 0x2801
 
 USE_LIMITED_API = (sys.platform != 'win32' or sys.version_info < (3, 0) or
-                   sys.version_info >= (3, 5))
+                   sys.version_info >= (3, 5)) \
+                  and not sysconfig.get_config_var("Py_GIL_DISABLED")
 
 
 class GlobalExpr:


arigo:
As described in #40 <https://github.com/python-cffi/cffi/issues/40> I we
would be quite surprised if cffi is stable out of the box with the new free
threading mode. This includes modules built with cffi. They may appear to
work, but knowing the source code I am rather sure that you have just been
lucky so far.

arigo:
Feel free to make a branch or fork for the free threading mode. I'm not
sure there is a point in adding a few easy fixes to the main branch if
there isn't a will to see the rest of the project through and extensively
test it.

minrk:
Sure. If you don't think it worth doing anything except fixing everything all in one PR, feel free to close, because that's not something I can tackle. Disabling py_limited_api is just required to even start exploring issues with cffi, and this is the only condition that seems to govern it.

Is the fact that py_limited_api=False doesn't actually disable the limited api a bug? I don't know enough to be sure.

arigo:
It certainly sounds strange.  But keyword arguments are passed straight to the underlying distutils/setuptools, so I don't actually know either.  As far as I know, `py_limited_api` wasn't even a valid keyword at the time the cffi code was written...  Feel free to add it as its own bug report?  I guess fixing that issue would also reduce the need for this PR for now...

minrk:
CFFI has [documentation](https://cffi.readthedocs.io/en/stable/whatsnew.html#v1-14-2) about supporting `py_limited_api=False`, which is what I tried first, before arriving at this PR when I discovered the documented argument has no actual effect. There's no real need for this PR (even though I still think it's correct) after #106, since packages can start experimenting.

PR #111, opened 2024-08-12
Use '_cffi_float_complex_t' or '_cffi_double_complex_t' more systematically

arigo:
Should fix the problem that FFI().typeof('float _Complex') stopped working.

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -217,7 +217,7 @@
 #define CT_UNION              0x080   /* union */
 #define CT_FUNCTIONPTR        0x100   /* pointer to function */
 #define CT_VOID               0x200   /* void */
-#define CT_PRIMITIVE_COMPLEX  0x400   /* float _Complex, double _Complex */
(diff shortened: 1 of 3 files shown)


testing/cffi1/test_parse_c_type.py
```diff
@@ -164,6 +164,8 @@ def test_simple():
             ("long double", lib._CFFI_PRIM_LONGDOUBLE),
             (" float  _Complex", lib._CFFI_PRIM_FLOATCOMPLEX),
             ("double _Complex ", lib._CFFI_PRIM_DOUBLECOMPLEX),
+            ("_cffi_float_complex_t", lib._CFFI_PRIM_FLOATCOMPLEX),
+            (" _cffi_double_complex_t", lib._CFFI_PRIM_DOUBLECOMPLEX),
```
nitzmahone, line 168: Is the leading space intentional?

arigo: Yes, absolutely.  It's a test, and there are already random spaces in the lines before.

PR #112, opened 2024-08-19
MNT: bump version number in setup.py

tacaswell:
Otherwise a source install sorts as older than 1.17.0 from pypi.


arigo:
Maybe call it 17.0.1dev0?  Otherwise we can't release any other 17.x.x without making it appear older than previous checkins to the 18.0.0dev0 source.

tacaswell:
Done!

arigo:
Some version numbers were not updated.  Please run `test_version` or look in the failures of the automated test runs.

tacaswell:
`__init__.py`` had both a `__version__` (which I caught) and a `__version_info__` (which I did not).

Should be fixed now and the tests pass locally.

PR #113, opened 2024-08-20
Remove now-obsolete NativeIO shim.

inklesspen:
This was only needed for Python 2 compatibility, which was dropped in v1.16.0.

### DIFF
--- a/src/cffi/recompiler.py
+++ b/src/cffi/recompiler.py
@@ -1410,15 +1410,6 @@ class Recompiler:
         self.cffi_types[index] = CffiOp(OP_ENUM, enum_index)
 
 
-if sys.version_info >= (3,):
-    NativeIO = io.StringIO
-else:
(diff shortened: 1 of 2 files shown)


arigo:
I'm not too keen on this kind of breaking change with no real purpose.  It would be cool if I, for example, could still download the latest version and run it in python 2.7 with minimal fixes, and I know other people like me.

inklesspen:
Can you run it on 2.7, though? setup.py says the minimum python version is 3.8, and has done since v1.16.0

inklesspen:
Huh, I guess it _does_ work on 2.7. Pip won't install it, but if you download the tarball and run setup.py it will happily install (although `cffi/_shimmed_dist_utils.py` uses some syntax that Python 2.7 doesn't support, when warning about Python 3.12 issues.)

Issue #47, opened 2024-01-11
Document how to use meson to build a cffi cextension


mattip:
setuptools is not the only python build backend. Projects in the scientific python stack have moved to [meson](https://mesonbuild.com/index.html). It would be nice to provide an example of how to best use meson to build a cffi c-extension module.

mattip:
I think this would help @minrk in ported pyzmq to meson.

inklesspen:
I believe the big issue with this is that cffi relies on setuptools in some cases; specifically, it uses the distutils Distribution and Extension classes to compile extension modules, and as of Python 3.12 these have been removed from the stdlib and are only available in setuptools.

However, distutils is only actually needed for the step which automatically compiles the generated c source. Unfortunately, at the moment doing without this requires digging a little deeper into the internals of cffi. (I suspect it could be made simpler to access, but I don't know what constraints the cffi devs are working under here.)

```python
from cffi import FFI

ffibuilder = FFI()

ffibuilder.cdef("int abs(int);")

ffibuilder.set_source("_simple", "#include <stdlib.h>")

if __name__ == "__main__":
    from cffi.recompiler import Recompiler
    # This code is based on FFI.emit_c_code and the functions it calls.
    # kwds are options which would be passed to the distutils Extension class
    # see https://setuptools.pypa.io/en/stable/userguide/ext_modules.html#setuptools.Extension
    # this may or may not be relevant when using meson
    module_name, source, source_extension, kwds = ffibuilder._assigned_source
    recompiler = Recompiler(ffibuilder, module_name)
    recompiler.collect_type_table()
    recompiler.collect_step_tables()
    filename = module_name + source_extension
    with open(filename, "w") as f:
        recompiler.write_source_to_f(f, source)
```

This Python script can be run with just `cffi` as a dependency. It generates the C extension source and writes it to a file; it should be obvious how to make it write to a different location. I suspect you can make this work with meson's [Generating sources](https://mesonbuild.com/Generating-sources.html) support.

arigo:
@inklesspen This might be an oversight.  Would it help if we move the line 1546 of ``recompile.py`` into the ``if call_c_compiler:`` branch a few lines below?

```
diff --git a/src/cffi/recompiler.py b/src/cffi/recompiler.py
index 4167bc05..19522470 100644
--- a/src/cffi/recompiler.py
+++ b/src/cffi/recompiler.py
[...]

PR #115, opened 2024-08-20
Allow writing generated code to a file-like object.

inklesspen:
This allows a caller to better control when and where the generated code is written, for issue #47.


inklesspen:
Admittedly, I'm guessing at the version number where this will be released. But that's simple enough to change.

inklesspen:
@arigo Hi, do I need to do something to kick off the workflow here?

arigo:
Sorry, I guess I missed the creation of this issue.

PR #123, opened 2024-09-05
Remove ResourceWarnings from tests

rtobar:
The open().read() calls were causing ResourceWarnings because the file was never explicitly closed, and are now replaced with a context managed open() call. In the case of the popens the reading from the stdout member had the same issue, so I'm replacing the stdout.read() + wait() with communicate() (which internally exhausts stdout and waits) + returncode.

All in all, this takes the warning count when running all tests from 45 down to 11, all of which are PytestUnraisableExceptionWarnings.


arigo:
LGTM, thanks!

PR #124, opened 2024-09-05
Remove pre-Python 3.8 C code

rtobar:
This is a bit of a big one, but it's mostly removal of dead code anyway, so hopefully it's not too controversial, and well received.

While doing some further code inspection today to re-tackle #86 I realised that there's still a fair bit of code in the C extension, and in the C extension generator, for Python < 3.8. Since 3.8 is the minimum supported version, it makes sense to remove that now-dead code in favour of a less complex codebase.

In order to make this a bit more digestible I split this work in two commits: one focuses on removing support for Python <3, and another for <3.8. Many of these changes are big `#if` blocks that could easily be removed. In a few places there were macros that were defined differently depending on the Python version, that would have now been left with a single definition. In these cases I removed the macro altogether in favour of using the direct Python C API or idiom as required (e.g., all the `PyText_*` macros are now `PyUnicode_*` calls, etc).

Both commits are (I think) correctly self-contained, at least locally they individually both compile, install and test correctly. So in case something breaks it should be easy to bisect back.

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -141,18 +141,8 @@
 #include "malloc_closure.h"
 
 
-#if PY_VERSION_HEX >= 0x03030000
-# define PyText_GetSize PyUnicode_GetLength
-#else
-# define PyText_GetSize PyUnicode_GetSize
-#endif
-
 #if PY_VERSION_HEX < 0x030900a4
 # define Py_SET_REFCNT(obj, val) (Py_REFCNT(obj) = (val))
-#endif
-
-#if PY_VERSION_HEX >= 0x03080000
-# define USE_WRITEUNRAISABLEMSG
 #endif
 
 /************************************************************/
@@ -4431,7 +4421,7 @@ static void *b_do_dlopen(PyObject *args, const char **p_printable_filename,
             if (*p_printable_filename == NULL)
                 return NULL;
 
-            sz1 = PyText_GetSize(filename_unicode) + 1;
+            sz1 = PyUnicode_GetLength(filename_unicode) + 1;
             sz1 *= 2;   /* should not be needed, but you never know */
             w1 = alloca(sizeof(wchar_t) * sz1);
             sz1 = PyUnicode_AsWideChar(filename_unicode,
@@ -5137,7 +5127,7 @@ static PyObject *b_complete_struct_or_union(PyObject *self, PyObject *args)
         if (!(sflags & SF_GCC_ARM_BITFIELDS) && fbitsize >= 0) {
             if (!(sflags & SF_MSVC_BITFIELDS)) {
                 /* GCC: anonymous bitfields (of any size) don't cause alignment */
-                do_align = PyText_GetSize(fname) > 0;
+                do_align = PyUnicode_GetLength(fname) > 0;
             }
             else {
                 /* MSVC: zero-sized bitfields don't cause alignment */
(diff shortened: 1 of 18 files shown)


arigo:
See https://github.com/python-cffi/cffi/pull/113 for my personal stance on this issue.  @nitzmahone If wants to merge this anyway, go ahead.

rtobar:
@arigo thanks for that input. I also maintain two libraries for which I kept 2.7 support until not too long ago, but finally decided to drop it, with the rationale being that 2.7 users still can use the old versions without lacking much functionality anyway. Kind-of-supporting 2.7 (or anything other than what `setup.py` states) will add a maintenance cost that officially you shouldn't need to pay.

I'm not sure if you're aware of this, but have a look at https://pypistats.org/packages/cffi. Seeing those numbers was also something that prompted me to drop support for <1% of users, who again can still download earlier versions of my library. Removing hundreds of lines of C code and macros I think pays the price.

My opinion only of course, if you decide you want to keep the code that's fine, you know your audience and needs better, I won't be pushing back.

nitzmahone:
Yeah, especially with the very deep inspection and review of all the code that's going to be necessary for *proper* support of free-threaded Python builds, I'm all for minimizing compatibility code in the active development branches for Python versions that aren't supported/tested.

My experience has repeatedly taught me: if it's not getting tested, it's probably broken...

The old branches of course aren't going anywhere, so if someone *really* needs to make something work on a long-dead Python version, they can always install an older release or backport fixes to a private fork. We've had several firedrills in recent memory caused by "dead but still called" code breaking suddenly, so I'm quite happy to be more aggressive about cleaning that stuff up for 1.18 and beyond. Thanks!

rtobar:
Thanks @nitzmahone, that sounds pretty good to me :)

I have been building some further changes on top of these in the hopes that they'd be merged. Based on your message I'll continue under that assumption.

Would you be happy for a similar effort in the python codebase?

rtobar:
Gentle ping here to see if the intention of doing this clean-up is still there, thanks!

rtobar:
Another gentle ping after a few months of inactivity

nitzmahone:
> I think we should release 1.18 as the last version to still support Python2, and merge something like this PR together with the free-threading code for a new version (1.19? 2.0?).

Is there actually anything unreleased or pending merge to `main` that's apropos for the older pypy branches? If not, the 1.17 branch is still there (and worst case, we could always backport whatever for a 1.17.2 in the future). IIUC the desire for a published pre-release with free-threaded support is the only time-sensitive item right now, so it'd be nice to avoid the churn of doing a 1.18 with nearly no changes and 1.19 with FT support concurrently.

mattip:
> it'd be nice to avoid the churn of doing a 1.18

OK, this is reasonable.

> could their "current CFFI" checks be retargeted to the release-1.17 branch

Good idea.

> It would be nice if removing-Python-2-support and free-threading-support would both be released into cffi 2.0.

We could do 2.0 with free-threaded support and partial Python2 cleanup (i.e. what is needed to make the merge of #178 easier) while declaring it no longer supports Python2. Then we could continue the cleanup in a 2.1. This would allow a faster release of the 2.0 free-threading support.

mattip:
I would prefer to separate the two commits. If [these statistics](https://pypistats.org/packages/urllib3) can be believed, 3.8 still has about a 10% share on urllib3 downloads on PyPI

mattip:
… although in the pyproject.toml there is already this which came in PR #129: https://github.com/python-cffi/cffi/blob/67a170dc7972dc9361743690b38da3d963b200ef/pyproject.toml#L15

@nitzmahone was that intentional?

mattip:
2.0 has been released. There are conflicts here, it would be nice to clean this up.

rtobar:
@mattip happy to, as long as there is a good level of certainty that these changes will be merged. They've been floating for 1.5 years now

PR #125, opened 2024-09-06
Hard disable Py_LIMITED_API under free-threaded build

nitzmahone:
* `Py_LIMITED_API` is not currently supported on the `t` ABI- it will build successfully, but segfaults. Hard disable for free-threaded Pythons until it's working.
* full test suite passes under free-threaded Linux x86_64 as of 3.13.0rc2
* update some broken test assumptions around limited API support and eager global GC
* add non-artifact CI job for 3.13t

NB: CI will fail on this job until 3.13.0rc2 is available in manylinux and cibuildwheel updates to include it- rc2 fixes [embedded interpreter init from non-main threads](https://github.com/python/cpython/issues/123022), which causes several of the embedding tests to fail.


arigo:
Sorry, you're on your own here.  I don't have time for a proper in-depth review, and I don't want to do a half-serious one.

nitzmahone:
@arigo No worries- I'll have @mattclay review before I merge (we wrote part of this together this afternoon anyway just messing around).

In general, is there any area or functionality where you definitely *would* want an opportunity to review changes, or should we just merge stuff once we're happy with it and it's passing tests?

arigo:
I wish I was still interested in the Python ecosystem enough for the free-threading challenge, but it's just not the case.  But feel free to ask and tentatively assign me for reviews, of course!  But as far as I can see, most of the rest that needs to be done is just tweaks for the new version of CPython/setuptools/gcc/whatever.

mattclay:
It's worth mentioning that supporting free-threaded builds does not automatically mean running without the GIL. An extension must also opt-in to running without the GIL [1]. Otherwise the GIL is automatically enabled at runtime when the extension is loaded and a warning is issued:

> The global interpreter lock (GIL) has been enabled to load module '_cffi_backend', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.

[1] https://docs.python.org/3.13/howto/free-threading-extensions.html#single-phase-initialization

mattclay (approved):

PR #127, opened 2024-09-06
[1.17 backport] Use '_cffi_float_complex_t' or '_cffi_double_complex_t' more systematically (#111)

nitzmahone:
Should fix the problem that FFI().typeof('float _Complex') stopped working.

* test fixes

(cherry picked from commit ac407f073068041a5ea9874d8b17f364865bf79b)


jhale:
Would it be possible to release this as 1.17.2? It's a blocker for us without pinning.

PR #135, opened 2024-10-11
rename possibly identical modules to prevent test collisions

### LINKED ISSUE #134: Running tests in parallel can cause compilation errors
When running tests with pytest-xtest, different test runs will try to compile the test code in the same `__pycache__` directory. If the tests contain the same C code, this will lead to a situation where two tests are trying to build the same extension simultaneously, which can lead to compilation overwriting the same files, which will not work on windows (and perhaps elsewhere). 

For instance, [these tests](https://github.com/python-cffi/cffi/blob/88f48d22484586d48079fc8780241dfdfa3379c8/testing/cffi0/test_verify.py#L129) will produce the same cffi source hash so calling `ffi.verify` will generate exactly the same object and linker output
```
def test_rounding_1():
    ffi = FFI()
    ffi.cdef("double sinf(float x);")
    lib = ffi.verify('#include <math.h>', libraries=lib_m)
    res = lib.sinf(1.23)
    assert res != math.sin(1.23)     # not exact, because of double->float
    assert abs(res - math.sin(1.23)) < 1E-5

def test_rounding_2():
    ffi = FFI()
    ffi.cdef("double sin(float x);")
    lib = ffi.verify('#include <math.h>', libraries=lib_m)
    res = lib.sin(1.23)
    assert res != math.sin(1.23)     # not exact, because of double->float
    assert abs(res - math.sin(1.23)) < 1E-5
```

One possible mitigation could be to change the `__init__` of the `FFI` class to add a random `ffi.cdef("/* %d */" % random.randint(0, 32_000))

mattip:
Fixes #134 by renaming the test module, if not already specifically named. I needed two `randint` invocations to reduce the chances of collision.


arigo:
I don't know what to do with the failures shown here on macosx.  I'm not sure if they are relevant, and if not, I'm not sure how to say "merge anyway please".

arigo:
Also, not sure why you use `randint(32000)`; I would replace that with `randint(1000000000)` and still call it twice.  It costs nothing and reduces the risks more.

mattip:
OK. I can reproduce the failures locally. [That test](https://github.com/python-cffi/cffi/blob/88f48d22484586d48079fc8780241dfdfa3379c8/testing/cffi0/test_verify.py#L2159) seems specifically designed to reuse the module, note how the lib2 code is exactly the same as the original one.

arigo:
Still fails.  Maybe differently?

arigo:
Still failing...

arigo:
Yes, that's bad.  I don't remember any failure caught specifically by `test_vgen2`, but its point was precisely to reuse the modules.

arigo:
This dooms the whole approach, I'm afraid.  We need either to say `pytest-xtest` is not supported, or decide `test_vgen2` is not very useful and kill it, or maybe find a proper refactoring.

mattip:
Is it important that __all__ the `test_vgen` tests be rerun without `*.c` sources, or can I add a specific test for a single case of removing the `*.c` source and running `verify` again, and actually check that the source is not regenerated? As far as I can tell, `test_vgen2` does not actually check that the source is unneeded, so I am not sure what it is designed to check.

mattip:
The latest commit
- removes the `test_v*2.py` files and adds a `test_no_regen` that checks the C file is not regenerated
- moves the global `verifier._TMPDIR` and `verifier._FORCE_GENERIC_ENGINE` into a `threading.local()` to make the global state more thread-safe
- refactors compilation to lessen the likelyhood of creating a filename that is longer than MAXPATH on windows.

mattip:
The latest commit
- uses filelock to make test_parse_c_type build lib only once
- restores `verifier.sourcefilename` after shortening it for compilation
- documents the need for filelock
- runs tests with `pytest-xdist -n 4` and add back windows testing

mattip:
Adding back windows testing almost worked, I had to skip embedded tests. I will open an issue so if someone is so inclined they can fix it. The tests are failing on the PyPy windows buildbot as well.

doc/source/installation.rst
```diff
@@ -54,8 +55,7 @@ Download and Installation:
   ``git clone https://github.com/python-cffi/cffi``
 
 * running the tests: ``pytest  c/  testing/`` (if you didn't
-  install cffi yet, you need first ``python setup_base.py build_ext -f
-  -i``)
+  install cffi yet, you need first ``python -m pip install -e .``)
```
mattip, line 58: These days `pip install` is to be preferred

testing/cffi0/test_verify.py
```diff
@@ -126,22 +137,14 @@ def test_Wconversion_double2int():
...
     assert abs(res - math.sin(1.23)) < 1E-5
 
-def test_rounding_2():
-    ffi = FFI()
-    ffi.cdef("double sin(float x);")
-    lib = ffi.verify('#include <math.h>', libraries=lib_m)
-    res = lib.sin(1.23)
-    assert res != math.sin(1.23)     # not exact, because of double->float
-    assert abs(res - math.sin(1.23)) < 1E-5
-
```
mattip, line 144: This test is the same as the one above it

mattip:
CI is passing.

testing/cffi0/test_verify.py
arigo: Er, no?  One is testing the 'sin' and the other the 'sinf' function.

mattip: Ok, restoring

arigo:
I can approve this without looking in details as it mostly touches tests.  Can you confirm the underlying reason, though: why do we want to run pytest-xdist?  The negative points are that it creates issues and adds extra installation dependencies; what is the positive point?

mattip:
On windows, tests now run in CI in ~11 minutes, where before they were 45 minutes.

arigo:
It would be very nice, if cffi was actively being developed.  Bu as it is not, my point stands..?

mattip:
these tests are run by pypy (in its extra tests) for every nightly buildbot run. The windows runs are very slow. So these changes are more for the PyPy test than for cffi itself, trying to minimize the diff between the two copies. But I see your point and will close this.

arigo:
Oh, if it makes the nightly test runs of pypy more than 30 minutes faster, that's an interesting outcome.

arigo:
Should I review this?  Or wait for #137 (which apparently needs some fixing)?

Issue #139, opened 2024-10-29
old-style callback in Windows worked with cffi 1.16, crashes with cffi 1.17.1. 


jmp75:
## Context

A colleague reported an issue that I first reproduced as a [0xC0000005: Access violation executing location on some windows platforms](https://github.com/csiro-hydroinformatics/moirai/issues/1). After a bit of diagnosing I noticed a change of implementation of `dlopen` for Windows in `cffi=1.17`, and the crash does not occur with `cffi=1.16`.

Also relates to [this google group post](https://groups.google.com/g/python-cffi/c/OnL09z8HNs8)

## Repro

A repro is available from [jmp75/py-cffi-callback-repro @ f7e1267](https://github.com/jmp75/py-cffi-callback-repro/tree/f7e12674b688aaee504fe360676000948c48246c). The program completes with `cffi==1.16`, but crashes with `cffi==1.17.1` when activating the callback:

```text
before registration
after registration
before has_callback_registered
after has_callback_registered
before triggering callback
...crash...
```

## Observations

The change in the underlying implementation of dlopen for Windows may have been motivated by [cffi/issues/64](https://github.com/python-cffi/cffi/issues/64) , leading to the change [Win32: pass the flags from dlopen() to LoadLibraryEx() #65](https://github.com/python-cffi/cffi/pull/65/commits/9c2dfe8aaa6fdd455bea4bb6b22f6be5dc12cabd).

`dlopen` under the hood in windows used to call `return (void *)LoadLibraryA(filename);` and it has now changed to `return (void *)LoadLibraryExA(filename, NULL, flags);`

I note I always passed `flag(s) = 1`, and note it seems to have been previously ignored on Windows.

mattip:
> I note I always `passed flag(s) = 1`

[These lines](https://github.com/jmp75/py-cffi-callback-repro/blob/f7e12674b688aaee504fe360676000948c48246c/native_handle.py#L43) seem wrong for windows 

```
flags = int(os.environ.get("RT_CODE", 1))
ut_dll = ut_ffi.dlopen(str(native_lib_path), flags)  # Lazy loading
```

Looking at the meaning of 1 in the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexa) it says 
[...]

PR #140, opened 2024-10-29
tweak the logic for LoadLibraryEx() calls following ctypes

mattip:
From [this comment](https://github.com/python-cffi/cffi/issues/139#issuecomment-2443489628)

@arigo 

Does it need any new test?

### DIFF
--- a/src/c/misc_win32.h
+++ b/src/c/misc_win32.h
@@ -191,13 +191,19 @@ static PyObject *b_getwinerror(PyObject *self, PyObject *args, PyObject *kwds)
 #define RTLD_GLOBAL 0
 #define RTLD_LOCAL  0
 
+#define MAX_BUF_FULLPATH  32768
+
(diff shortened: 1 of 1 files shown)


arigo:
I'm not even absolutely sure about why the ctypes logic is that way or the exact behavior that it implies.  As far as I can tell the code here is doing the same thing, but not sure I can write relevant tests.

arigo:
We can write a comment for the whatsnew and/or the documentation: when `ffi.dlopen(filename, flags)` is called on Windows with flags equal to zero or omitted, then:

1. cffi automatically provides some reasonable flags to `LoadLibraryEx()`;
2. if `filename` has got a path, but a relative one, then it is replaced with an absolute path.

If needed, you can avoid all this by giving an explicit non-zero flags.

arigo:
Is there still something to do here?

PR #143, opened 2024-11-13
Make ffi_init_once thread-safe in free threaded build

colesbury:
Lock the FFI object when initializing the init_once_cache dictionary and use PyDict_GetItemRef in places where the dictionary's value may be concurrently modified.

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -204,6 +204,27 @@
 # define USE_WRITEUNRAISABLEMSG
 #endif
 
+#if PY_VERSION_HEX <= 0x030d00a1
+static int PyDict_GetItemRef(PyObject *mp, PyObject *key, PyObject **result)
+{
+    PyObject *obj = PyDict_GetItemWithError(mp, key);
+    Py_XINCREF(obj);
+    *result = obj;
+    if (obj == NULL) {
+        if (PyErr_Occurred()) {
+            return -1;
+        }
+        return 0;
+    }
+    return 1;
+}
+#endif
+
+#if PY_VERSION_HEX <= 0x030d00b3
+# define Py_BEGIN_CRITICAL_SECTION(op) {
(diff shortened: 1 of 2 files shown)


colesbury:
Hi @nitzmahone - I'm still working through the CFFI free-threading issues, but I figured it'd be helpful to get some feedback on a small change.

On that note:
* I added a "polyfill" for `PyDict_GetItemRef`, but that sort of thing is nicely provided. by https://github.com/python/pythoncapi-compat/. Would you be interested in using that?
* It's probably cleaner and simpler to make `init_once_cache` have the same lifetime as its FFIObject instead of lazily initializing it. I went with the smaller code change for now, but let me know what you think.

To find thread-safety issues, in addition to reading through the code, I'm using https://github.com/Quansight-Labs/pytest-run-parallel, which I've found to be very effective.

colesbury:
Hey @nitzmahone - gentle ping. I'd really like to see python-cffi work in the free threading build as a lot of libraries depend on it. I'm happy to do most of the implementation work.

arigo:
Hi!  Just in case you want an alternative approach, you can make a fork of cffi and implement and test free-threading in it.  Ideally you'd advertise it for pip usage too (I think it can be done with a direct github url if you don't want to make a new cffi-xyz package name).  Once such a thing exists, is complete, and is reasonably well-tested, I'd be up to reviewing a single big pull request.  What I won't do is review many small incremental patches---but I'm not speaking for nitzmahone here, only for me.

colesbury:
Hi Armin - thanks for your suggestion and I appreciate your offer to review a single, big, well-tested PR.

A forked, free threading pacakge is a bit awkward because pip doesn't support "free threading" as a constraint: a project can't easily depend on a forked version only for the free threading build. I think there are workarounds, but they require more substantial changes to the build process of packages that depend on cffi.

ngoldbaum:
This can be closed in favor of #178.

Issue #147, opened 2024-12-08
[ffi.compile] Sizes are not properly resolved when a referenced type X has a pointer to type which includes X


droptopx:
The following files cause a crash due to

`TypeError: field 'struct Outer.inner' has ctype 'struct Inner' of unknown size`

when they are run in sequence:

`gen_cffi_test.py`:
```python
import cffi

ffi = cffi.FFI()
ffi.set_source("bug_repro", None)
ffi.cdef(
    """\
struct Inner {
  struct Outer *outer_ref;
};

struct Outer {
  struct Inner inner;
};
"""
)
ffi.compile(verbose=True, debug=True, tmpdir="./bug_repro_autogen")
```

`cffi_test.py`:
```python
from bug_repro_autogen.bug_repro import ffi

ffi.callback("void (struct Inner *)")
```

`python gen_cffi_test.py && python cffi_test.py`

I have not figured out what the root cause is and will probably not be able to find time to look further. Any analysis by someone else is welcome.

EDIT: It's not just `ffi.callback`, `ffi.sizeof("struct Inner")` also gives the same error.
EDIT2: The error doesn't happen when the erroring method is called in the first file (i.e. no compilation)

arigo:
It works for me and I remember fixing something like that (maybe last year?).  Can you check which version of CFFI you're testing?

arigo:
Sorry, I got it to fail too now.  My mistake.  It only occurs in out-of-line ABI mode, when you specify `None` for the C sources in `ffi.set_source()`.

arigo:
There is an order dependency, which suggests a workaround for you for now: if you first do `ffi.sizeof("struct Outer")` then any following usage of `struct Inner` works fine.

arigo:
See https://github.com/python-cffi/cffi/pull/148.

droptopx:
@arigo #148 does fix #147, thanks a lot for the fix and the great turn-around time!

PR #148, opened 2024-12-09
issue 147: force-compute nested structs before parent structs

arigo:
Occurs mainly in out-of-line ABI mode.

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -5238,6 +5238,13 @@ static PyObject *b_complete_struct_or_union(PyObject *self, PyObject *args)
                               &fbitsize, &foffset))
             goto error;
 
+        if ((ftype->ct_flags & (CT_STRUCT | CT_UNION)) &&
+            !(ftype->ct_flags & CT_IS_OPAQUE)) {
+            /* force now the type of the nested field */
+            if (force_lazy_struct(ftype) < 0)
+                return NULL;
+        }
+
         if (ftype->ct_size < 0) {
             if ((ftype->ct_flags & CT_ARRAY) && fbitsize < 0
                     && (i == nb_fields - 1 || foffset != -1)) {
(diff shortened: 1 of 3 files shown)


mattip:
Any idea how to port this to PyPy? The code there is different: there is no explicit `CT_IS_OPAQUE` flag. Even more complicated: I am not sure how to write a untranslated test for this.

arigo:
The PyPy logic is different, so maybe there isn't anything to fix.  Does the original test case in #147 work on a translated PyPy?

arigo:
Found the same logic in complete_struct_or_union() in newtype.py, but also found comments in the class W_CTypeStructOrUnion that make me think that PyPy, unlike CFFI on CPython, doesn't use `ftype->ct_size < 0` if the struct type is lazy.  If that's correct, then there is indeed nothing to fix.

mattip:
The test fails on py3.10. I tried various incantations to try to cause the struct to be evaluated when `size == -2`, which I think is supposed to be "lazy", but failed. I also could not figure out how to run the test untranslated.

arigo:
Here's a blind attempt at writing an untranslated test (completely untested):

```
diff -r 2eefecd93d0f pypy/module/_cffi_backend/test/test_re_python.py
--- a/pypy/module/_cffi_backend/test/test_re_python.py  Mon Oct 16 08:36:02 2023 +0200
+++ b/pypy/module/_cffi_backend/test/test_re_python.py  Sun Dec 29 18:54:07 2024 +0100
@@ -32,6 +32,7 @@
         struct foo_s;
         typedef struct bar_s { int x; signed char a[]; } bar_t;
         enum foo_e { AA, BB, CC };
+        struct B { struct C* c; }; struct C { struct B b; };

         void init_test_re_python(void) { }      /* windows hack */
         void PyInit__test_re_python(void) { }   /* windows hack */
@@ -270,3 +271,15 @@

         err = lib1.dlclose(handle)
         assert err == 0
+
+    def test_rec_structs_1(self):
+        self.fix_path()
+        from re_python_pysrc import ffi
+        sz = ffi.sizeof("struct B")
+        assert sz == ffi.sizeof("int *")
+
+    def test_rec_structs_2(self):
+        self.fix_path()
+        from re_python_pysrc import ffi
+        sz = ffi.sizeof("struct C")
+        assert sz == ffi.sizeof("int *")
```

mattip:
Yup, thanks @arigo that worked. The test and fix are in [this commit](https://github.com/pypy/pypy/commit/e904ef633470080ba2abdedc171743b91e387125)

PR #164, opened 2025-04-22
Build ARM64 Windows wheels

### LINKED ISSUE #109: Please provide Windows ARM64 wheels
There is now native ARM64 Python since release 3.12. many central projects including this one do not provide wheels for this platform. Please start building wheels also for Windows ARM64 when convenient.

lazka:
ARM64 Python is available since Python 3.11, so add specs for 3.11-3.13

Use actions/setup-python to avoid using the host Python (3.9 currently on windows-2022) since the arm image doesn't have a native Python atm (see https://github.com/actions/partner-runner-images/issues/85) and cibuildwheel warns that it will require 3.11 with v3, so might as well bump it for everything.

Also adds one arm64 build to the default matrix setup, since the separate image makes it different enough to warrant testing on PRs imo.

Fixes #109

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -454,44 +454,68 @@ jobs:
         matrix_yaml: |
           include:
           - spec: cp38-win_amd64
+            image: windows-2022
             omit: ${{ env.skip_ci_redundant_jobs }}
 
           - spec: cp39-win_amd64
+            image: windows-2022
             omit: ${{ env.skip_ci_redundant_jobs }}
 
           - spec: cp310-win_amd64
+            image: windows-2022
             omit: ${{ env.skip_ci_redundant_jobs }}
 
           - spec: cp311-win_amd64
+            image: windows-2022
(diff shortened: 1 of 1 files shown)


lazka:
Here is a full matrix build for this PR: https://github.com/lazka/cffi/actions/runs/14604224847

.github/workflows/ci.yaml
```diff
@@ -503,6 +527,11 @@ jobs:
       with:
         name: ${{ needs.python_sdist.outputs.sdist_artifact_name }}
 
+    - name: install python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.11
```
ofek, line 533: In order to postpone future required updates, what do you think about increasing this to the latest supported minor version?

lazka: sure, done

arigo:
@nitzmahone

khmyznikov:
@arigo  @nitzmahone @ofek What is the current blocker?

PR #165, opened 2025-04-28
No need to install wheel any more

stefanor:
setuptools has had a native bdist_wheel since 70.1.0: https://setuptools.pypa.io/en/latest/history.html#v70-1-0


stefanor:
It would probably be good to have some pep517 integration tests, those will have to use versions of `cffi` from PyPI, which makes it impossible to test any coupled changes.

mattip:
There are merge conflicts

stefanor:
Rebased.

PR #167, opened 2025-05-04
apply patch from issue 166

### LINKED ISSUE #166: MSVC error C2026: string too big
I have a huge enum (6000+ options) in my cdef. The generated C code for the array of `_cffi_enum_s`, field `enumerators` thus contains a humongous string which trips MSVC.

Huge generated strings need to be split. As a workaround I resorted to having `typedef ... MyEnum;` in the cdef, but now I don't have `MyEnum's` fields on my imported `lib` object.

mattip:
Closes #166

### DIFF
--- a/src/cffi/recompiler.py
+++ b/src/cffi/recompiler.py
@@ -93,9 +93,18 @@ class EnumExpr:
         self.allenums = allenums
 
     def as_c_expr(self):
-        return ('  { "%s", %d, _cffi_prim_int(%s, %s),\n'
-                '    "%s" },' % (self.name, self.type_index,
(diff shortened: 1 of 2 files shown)


mattip:
CI is passing. @gibki if you apply the patch to your CFFI does it fix the problem?

Lcbx:
Hello 👋
for what it's worth, I encountered the same issue and applying the patch fixed it.
(working on some procedural [steamworks API binding](https://github.com/Lcbx/steamworks_cffi/blob/65e372f1b03b94599f3dd5f35c6f4975270d876b/steamApi.py#L56) as a hobby project and there are huge enums)

PR #171, opened 2025-05-25
Change the license from MIT to MIT-no-attribution...

arigo:
…which is the same without the line that says "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

It should fix the "issue" reported here: https://groups.google.com/g/python-cffi/c/cbl-yazw3oE/m/2f8FCGYQFwAJ


mattip:
Thanks @arigo

PR #175, opened 2025-05-28
Use multi-phase init

AA-Turner:
#174.

cc @arigo -- this is a minimal implementation for `_cffi_backend.c`. Questions:

* Can we remove the `PY_VERSION` checks? It's hard to test the Python-2 case, and as far as I can tell you only support Python 3.8+.
* Linked to the above, ideally we would consider [module isolation](https://docs.python.org/3/howto/isolating-extensions.html), but this requires Python 3-only API that would mean more `#if` branching, so I've avoided for now.
* There's a use of single-phase init (`PyModule_Create`) in `vengine_cpy.py`, but the file is marked as deprecated. Is the file still used / should I update this?
* `_my_Py_InitModule()` and `b_init_cffi_1_0_external_module()` will be tricky to migrate. Would it be possible to use the [dynamic creation functions](https://docs.python.org/3/c-api/module.html#low-level-module-creation-functions) here instead?

A

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -8031,27 +8031,21 @@ static struct { const char *name; int value; } all_dlopen_flags[] = {
 
 /************************************************************/
 
-#if PY_MAJOR_VERSION >= 3
-static struct PyModuleDef FFIBackendModuleDef = {
-  PyModuleDef_HEAD_INIT,
-  "_cffi_backend",
-  NULL,
-  -1,
-  FFIBackendMethods,
-  NULL, NULL, NULL, NULL
-};
-#define INITERROR return NULL
+static int module_loaded = 0;
 
-PyMODINIT_FUNC
-PyInit__cffi_backend(void)
-#else
-#define INITERROR return
-
-PyMODINIT_FUNC
-init_cffi_backend(void)
-#endif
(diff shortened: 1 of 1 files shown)


mattip:
@ngoldbaum, is this part of #178 or are these PRs orthagonal?

mattip:
> Linked to the above, ideally we would consider [module isolation](https://docs.python.org/3/howto/isolating-extensions.html),

I think that is fine. Python changed the documentation: it used to state that mult-phase init implied module isolation, now it states that module isolation is recommended but not required.

mattip:
> There's a use of single-phase init (PyModule_Create) in vengine_cpy.py, but the file is marked as deprecated. Is the file still used / should I update this?

I see the engine choice logic is [here](https://github.com/python-cffi/cffi/blob/92645ec9329feb03e3fa86240b40fe0fc37469d1/src/cffi/verifier.py#L222). It seems `vengine_cpy.VCPythonEngine` is used if
- We are not on PyPy
- `ffi._backend` is not `_cffi_backend`. As far as I can tell this is only used in tests to try out the ctypes backend, and should not happen in "real life".

@arigo is that a correct assesment? If true, then I would not worry about `vengine_cpy` too much.

PR #184, opened 2025-07-29
Explicitly specify manylinux2014 in wheel building config

ngoldbaum:
While working on the wheel building CI for cryptography, I noticed that CFFI 2.0.0b1 is missing manylinux2014 wheels for some platforms. This is happening because cibuildwheel 3.0 changed the default manylinux image to manylinux_2_28. I think explicitly specifying manylinux2014 will create the manylinux2014 wheels that are currently missing.


mattip:
I don't see the artifacts uploaded.

ngoldbaum:
It looks like they're showing up attached to each job:

https://github.com/python-cffi/cffi/actions/runs/16601223055#summary-46961152640

PR #186, opened 2025-08-13
Allow riscv64 package

ffgan:
PyPI recently added support for RISC-V, so we might consider adding support for RISC-V64.

Below are the CI results:
[CFFI CI](https://github.com/ffgan/cffi/actions/runs/16931415744)

The results are generally well presented.

Other Info
Co-authored by: [nijincheng@iscas.ac.cn](mailto:nijincheng@iscas.ac.cn);

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -139,6 +139,13 @@ jobs:
           - { spec: cp314-manylinux_s390x, arch: s390x, omit: ${{ env.skip_slow_jobs }} }
           - { spec: cp314t-manylinux_s390x, arch: s390x, omit: ${{ env.skip_slow_jobs }} }
 
+          # riscv64 manylinux
+          - { spec: cp39-manylinux_riscv64, arch: riscv64, omit: ${{ env.skip_slow_jobs  || env.skip_ci_redundant_jobs}}}
+          - { spec: cp310-manylinux_riscv64, arch: riscv64, omit: ${{ env.skip_slow_jobs  || env.skip_ci_redundant_jobs}}}
+          - { spec: cp311-manylinux_riscv64, arch: riscv64, omit: ${{ env.skip_slow_jobs  || env.skip_ci_redundant_jobs}}}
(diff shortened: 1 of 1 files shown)


ffgan:
hi @ngoldbaum , Could you please help to check this PR? If there is anything that needs to be modified, please feel free to contact me.

mattip:
Could you point to a project that is releasing RISC-V wheels on PyPI? As far as I can tell, neither NumPy nor Cryptography have uploaded RISC-V wheels.

mattip:
Typically, we try to support earlier versions of `manylinux`. Currently we are producing manylinux2014_manylinux_2_17 wheels. The CI run you point to creates `manylinux_2_38_riscv64.manylinux_2_39_riscv64` wheels. Is that generally useful to the RISC-V community? Looking at the [pep 600 compatibility page](https://github.com/mayeut/pep600_compliance) it seems that supports openeuler 24.03 and later. But debian has support for RISC-V from 2.27, is that not commonly used?

markdryan:
> Typically, we try to support earlier versions of `manylinux`. Currently we are producing manylinux2014_manylinux_2_17 wheels. The CI run you point to creates `manylinux_2_38_riscv64.manylinux_2_39_riscv64` wheels. Is that generally useful to the RISC-V community? Looking at the [pep 600 compatibility page](https://github.com/mayeut/pep600_compliance) it seems that supports openeuler 24.03 and later. But debian has support for RISC-V from 2.27, is that not commonly used?

Almost all current manylinux images are derived from redhat compatible distros.  The manylinux_2_39_riscv64 image is based on RockyLinux 10 which was the first redhat compatible distro to produce a stable release for riscv64.  As such, we're unlikely to get an older manylinux image for riscv64 as there no suitable older distros available on which to base it.

manylinux_2_39_riscv64 wheels will run on Ubuntu 24.04 and RockyLinux10.  riscv64 users usually run fairly recent distros in order to take advantage of kernel and compiler support for newly ratified extensions.  For example, if you want to compile and run applications that use the RISC-V vector extension you need at least gcc-14 and a 6.9 kernel.

mattip:
Closing in favor of #227

Issue #187, opened 2025-08-19
Release CFFI 2.0.0


ngoldbaum:
I'm opening this now since 3.14.0rc2 came out last week. We should have one more release candidate in September and then the final release in October. It would be nice to have a CFFI release available sometime in September, since that will unblock work in reverse dependencies of cryptography and pynacl.

I'll defer to @nitzmahone on what the broader blockers on a final release are, but here's what I think needs to happen:

* [x] Ensure manylinux2014 wheels are built: #184 
* [x] Add a top-level thread safety page to the docs https://github.com/python-cffi/cffi/pull/188
* [x] Clean up the Windows testing situation. Fix [this commented out line](https://github.com/python-cffi/cffi/blob/b4bbe7940d3f76027534db1aecbae9b61c35221a/.github/workflows/ci.yaml#L357-L358) and make [this step](https://github.com/python-cffi/cffi/blob/b4bbe7940d3f76027534db1aecbae9b61c35221a/.github/workflows/ci.yaml#L431-L439) unconditionally run the same tests on all operating systems. #189 #137.

Anything else?

https://github.com/python-cffi/cffi/issues/126 alludes to a need for more multithreaded testing and better multithreaded test coverage. I'd appreciate more detail on what @nitzmahone had in mind there if he considers it a blocker.

I'd also like to make the `Py_mod_gil` setting configurable. Right now we unconditionally set it to `Py_MOD_GIL_NOT_USED`, which is true of the *wrapper*, but not necessarily the *wrapped* C code. This requires spitballing an API to control that at build time. I don't think this is a blocker since it can be worked around but I'm also happy to work on this now if you'd like.

I'm motivated to work on this but I'm going to need a little guidance on some of these and on whether there are any other blockers. For now I'm going to work on the thread safety docs and the windows testing situation.

nitzmahone:
I'd be thrilled if all of that was in for a September 2.0.0 - I probably won't have a ton of time to pitch in on implementation before then, but will definitely commit to reviews and release work as needed.

I had similar thoughts about coming up with a nicer way to expose a module init opt in/out for GIL behavior, but didn't want to hold up the initial pre-release trying to design it. IIRC I successfully hacked together a throwaway attempt with a completely manual call to it back in a 3.13 pre-release, but it feels like something that should be a little more declarative and less "everyone needs to reinvent this wheel".

[...]

PR #189, opened 2025-08-25
Enable more Windows pytest-run-parallel CI

ngoldbaum:
Towards #187.

This enables pytest-run-parallel CI on Windows for more tests. It also makes it so all platforms run the same tests under pytest-run-parallel.

Technically we're skipping tests now on Linux and Mac, but those tests are already being run in other CI jobs so no test coverage is lost.

### DIFF
--- a/testing/cffi1/test_function_args.py
+++ b/testing/cffi1/test_function_args.py
@@ -1,4 +1,9 @@
 import pytest, sys
+
+pytestmark = [
+    pytest.mark.thread_unsafe(reason="Workers would share a build directory"),
+]
+
 try:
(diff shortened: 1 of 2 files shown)


testing/cffi1/test_function_args.py
```diff
@@ -1,4 +1,9 @@
 import pytest, sys
+
+pytestmark = [
+    pytest.mark.thread_unsafe(reason="Workers would share a build directory"),
+]
```
ngoldbaum, line 5: this wasn't needed until now because this test file only runs on Windows, see below

PR #199, opened 2025-09-09
Fix license in pyproject.toml to be compatible with PEP 621

Bluefissure:
Update pyproject.toml to change the `license` field compatible with [PEP 621](https://peps.python.org/pep-0621/#license)

I'm not sure which is the best way to resolve this as the `license` is deprecated in a following [PEP 639](https://peps.python.org/pep-0639/) (which introduces `License-Expression`), but making this change will at least making new pip version (> 23.0.1) not complaining.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ requires-python = ">=3.9"
 
 description = "Foreign Function Interface for Python calling C code."
 readme = {file = "README.md", content-type = "text/markdown"}
-license = "MIT"
+license = { file = "LICENSE" }
 classifiers = [
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",


mattip (approved):
LGTM, pending CI completion. This is the same form of the `pyproject.toml` in scipy.

Bluefissure:
> I don't understand. We had a pyproject.toml that satisfied PEP 639. Then this PR undid this to return to a deprecated specification?

Yes, I think the [required `setuptools >= 66.1`](https://github.com/python-cffi/cffi/blob/bb50a2b430e4f31ec8cdcd5a7cb85da3fd524829/pyproject.toml#L5) does not support PEP 639, we need to update that if we want to change it to PEP 639.

Oh just saw your PR, thanks!

PR #215, opened 2025-11-22
Manually apply ruff/flake8-implicit-str-concat rule ISC002

DimitriPapadopoulos:
ISC002 Implicitly concatenated string literals over multiple lines


mattip:
Thanks @DimitriPapadopoulos

PR #214, opened 2025-11-22
Start removing pre-Python 3.8 code

DimitriPapadopoulos:
Companion to #124.

Mostly applied ruff [pyupgrade (UP)](https://docs.astral.sh/ruff/rules/#pyupgrade-up) rules, and some [pycodestyle (E)](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w) and [refurb (FURB)](https://docs.astral.sh/ruff/rules/#refurb-furb) rules.

**Edit:** I have left out some [pyupgrade (UP)](https://docs.astral.sh/ruff/rules/#pyupgrade-up) rules:
* [UP030](https://docs.astral.sh/ruff/rules/format-literals/), [UP031](https://docs.astral.sh/ruff/rules/printf-string-formatting/), [UP032](https://docs.astral.sh/ruff/rules/f-string/) are about moving string interpolation to f-strings. Change are pervasive, they are better left in a PR of their own.
* [UP036](https://docs.astral.sh/ruff/rules/outdated-version-block/) implies lots of manual changes and requires extra care. Again, best left in a PR of its own.

PR #218, opened 2025-12-10
Keep removing Python pre-3.9 code

DimitriPapadopoulos:
Companion to #124 and #214.

Apply rule [UP036](https://docs.astral.sh/ruff/rules/outdated-version-block/), it implies lots of manual changes and requires extra care.

~~Needs #214 to be applied first.~~

### DIFF
--- a/src/c/test_c.py
+++ b/src/c/test_c.py
@@ -69,17 +69,8 @@ import sys
 assert __version__ == "2.2.0.dev0", ("This test_c.py file is for testing a version"
                                      " of cffi that differs from the one that we"
                                      " get from 'import _cffi_backend'")
-type_or_class = "class"
-long = int
-unicode = str
-unichr = chr
-mandatory_b_prefix = 'b'
-mandatory_u_prefix = ''
-bytechr = lambda n: bytes([n])
-bitem2bchr = bytechr
-u = ""
+bitem2bchr = lambda n: bytes([n])
 str2bytes = lambda s: bytes(s, "ascii")
-strict_compare = True
 
 def size_of_int():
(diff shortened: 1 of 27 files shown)


src/c/test_c.py
```diff
@@ -2467,10 +2420,7 @@ def test_cmp():
     assert (p < s) ^ (p > s)
 
 def test_buffer():
-    try:
-        import __builtin__
-    except ImportError:
-        import builtins as __builtin__
+    import builtins
```
mattip, line 2406: I think this can be removed?

src/cffi/lock.py
```diff
@@ -1,15 +1,9 @@
...
-        from dummy_thread import allocate_lock
-else:
-    try:
-        from _thread import allocate_lock
-    except ImportError:
-        from _dummy_thread import allocate_lock
+try:
+    from _thread import allocate_lock
+except ImportError:
+    from _dummy_thread import allocate_lock
```
mattip, line 6: When might this fail? I think _thread is required for Python now.

mattip (commented):
I actually read the whole thing and only had two nits

src/cffi/lock.py
DimitriPapadopoulos: Indeed [`_thread`](https://docs.python.org/3/library/_thread.html#module-_thread)has always been part of Python 3.

MLopez-Ibanez:
This may need to be rebased and retested.

DimitriPapadopoulos:
Yes, this is still a work in progress. Some changes break CI. I need to apply smaller change sets gradually — after rebasing.

PR #221, opened 2025-12-12
Initialize interned_fields to NULL

### LINKED ISSUE #219: Code path leading to Py_XDECREF an unitialized  pointer
Hello,

Found via coverity scanning.

In https://github.com/python-cffi/cffi/blob/main/src/c/_cffi_backend.c

PyObject *interned_fields is declared: 

https://github.com/python-cffi/cffi/blob/d443106feb8faf4b6bef038bcb5cf821f3aa202b/src/c/_cffi_backend.c#L5155

goto finally; might be reached before the initialization on line 5335:

https://github.com/python-cffi/cffi/blob/d443106feb8faf4b6bef038bcb5cf821f3aa202b/src/c/_cffi_backend.c#L5171

Py_XDECREF(interned_fields); on a possible uninitialized value:

https://github.com/python-cffi/cffi/blob/d443106feb8faf4b6bef038bcb5cf821f3aa202b/src/c/_cffi_backend.c#L5510

stratakis:
This avoids a code path where Py_XDECREF could be used on interned_fields when uninitialized.

Resolves #219

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -5152,7 +5152,7 @@ static PyObject *b_complete_struct_or_union_lock_held(CTypeDescrObject *ct,
     Py_ssize_t byteoffsetorg;
     CFieldObject **previous;
     int prev_bitfield_size, prev_bitfield_free;
-    PyObject *interned_fields;
+    PyObject *interned_fields = NULL;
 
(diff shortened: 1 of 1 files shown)


mattip:
Thanks @stratakis

PR #225, opened 2026-02-04
explicitly add string.h and stdlib.h headers

kumaraditya303:
Under newer limited C API versions, these headers are not included by default so this PR adds them explicitly to allow compilation.


mattip:
Thanks @kumaraditya303

PR #227, opened 2026-02-05
ci: add riscv64 manylinux wheels

justeph:
Now that cibuildwheel and PyPI support riscv64, we can start building riscv64 wheels.

Integrated in the same way as the s390x platform.


mattip:
Why didn't CI run here, or did it run and the results are no longer around?

mattip:
Close open to trigger CI

mattip:
Thanks @justeph, @ffgan.

PR #235, opened 2026-03-20
do not crash on __delitem__

mildsunrise:
what it says on the tin: we're forgetting to check for deletions in many cases, so it's easy to crash the python process by doing `del pointer[slice]` or `del buffer[anything]`


mattip:
Please add a test that fails before, passes after

mildsunrise:
done. i don't think it's possible to test in a way that fails instead of invoking UB... in this case i think it's mostly guaranteed to be a crash though

mildsunrise:
hello?

mattip:
Thanks @mildsunrise

PR #239, opened 2026-03-25
Introducing macos-26

DimitriPapadopoulos:
I understand `pytest-run-parallel` should run on the latest runners available for each OS.


mattip:
Thanks @DimitriPapadopoulos

PR #240, opened 2026-04-08
Build Android wheels in CI

timrid:
This PR adds building wheels for Android in CI. 

Android is official supported since 3.13 (see [PEP 738](https://peps.python.org/pep-0738/)), so only wheels for Python 3.13 and 3.14 are build.

The prebuild libffi from https://github.com/beeware/cpython-android-source-deps is used similar to the [iOS build](https://github.com/python-cffi/cffi/pull/181). This prebuild version is the same that the official python build uses (see [here]( https://github.com/python/cpython/blob/a84e2dba24286d9cf13c29807d6994a13dc102a1/Android/android.py#L215)).

FYI: @mhsmith

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -481,8 +481,104 @@ jobs:
         if-no-files-found: error
       if: ${{ env.skip_artifact_upload != 'true' }}
 
+  make_android_matrix:
+    runs-on: ubuntu-24.04
+    outputs:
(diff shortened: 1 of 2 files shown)


mhsmith:
Hi, I'm the maintainer of Python's Android support. Could someone take a look at this PR when you have time? cffi is one of the most popular [packages not yet supported on Android](https://beeware.org/mobile-wheels/), and there are many other things that depend on it.

Thanks.

PR #76, opened 2024-05-15
Add "buildtool" to generate C source for cffi-based modules.

inklesspen:
This is useful when using alternative PEP 517 build systems, such as meson-python. It can be used as a CLI script or from an `if __name__ == "__main__"` block.

Examples of usage:

* CLI tool: `cffi-buildtool _fontconfig.c.py _fontconfig.c`
* Called from build script:
```python
if __name__ == "__main__":
    from cffi.buildtool import generate_c_source
    print(generate_c_source(ffibuilder))
```

This should be useful in a variety of build scenarios, not just supporting meson-python.

Fixes #47.

However, I think I should probably add some tests, and I do not understand the structure of the CFFI tests. Please advise.

Also I want to write some docs explaining how to use it, but I will do that after you give the approval for this basic approach.

PR #241, opened 2026-04-20
Integrate cffi-buildtool into CFFI itself

### LINKED ISSUE #47: Document how to use meson to build a cffi cextension

ngoldbaum:
Closes #47. Revives #76.

This adds a new submodule for cffi named `cffi.buildtool` and a new CLI script distributed by CFFI named `gen-cffi-src`. Also adds new tests and documentation. See the new documentation and tests for usage details.

I realize that this is a big change that makes a number of opinionated choices. I'm very happy to adjust things to suit the maintainers' taste.

My ultimate goal here is to make it possible for projects using CFFI to use arbitrary build tools and remove the perceived tight coupling between CFFI and setuptools/distutils. If there isn't appetite to add the code in cffi-buildtool to CFFI itself, I'd also happily document how to depend on and use cffi-buildtool instead.

See https://github.com/inklesspen/cffi-buildtool/issues/2 where I got @inklesspen's blessing to do this.

The is based on code and documentation that were originally written by Rose Davidson (@inklesspen on GitHub) for the cffi-buildtool project: https://github.com/inklesspen/cffi-buildtool.


inklesspen:
Just commenting here to confirm I give my full support for this upstreaming, and after a release is cut which includes these changes, I will mark my existing project as deprecated.

Also I'm available to answer any questions about the original code upon which this is based.

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+
+.. contents::
+
+CFFI ships a small subpackage, :mod:`cffi.buildtool`, together with a
+command-line program, ``gen-cffi-src``. Both produce the same output as
+:meth:`FFI.emit_c_code`: a ``.c`` source file ready to be compiled into
+a CPython extension module. What they add is two convenient front-ends
+-- one that executes an existing "build" Python script, and one that
+reads a ``cdef`` and C prelude from two files. This tool enables
+integrating with and build backend, such as `meson-python
```
ngoldbaum, line 15:
```suggestion
integrating with any build backend, such as `meson-python
```

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+
+   $ gen-cffi-src read-sources squared._squared squared.cdef.txt squared.csrc.c _squared.c
+
+The first positional argument is the fully qualified module name that
+will be embedded in the generated source (equivalent to the first
+argument to :meth:`FFI.set_source`).
+
+
+A Worked Example Using ``meson-python``
+=======================================
```
ngoldbaum, line 144: Maybe this should just link to the example projects in the tests to avoid duplication?

testing/cffi1/buildtool_example2/src/csrc/square.c
ngoldbaum, line 1: The files in this second example that are exact duplicates of the ones in the first example could probably be symlinks.

ngoldbaum (commented):
Some thoughts that occurred to me since yesterday and a typo fix

doc/source/buildtool.rst
inklesspen: I think there's value in having it right there in the docs, personally.

testing/cffi1/buildtool_example2/src/csrc/square.c
inklesspen: Well, in my version of the project I used an actual (though very simple) [C library](https://github.com/inklesspen/cffi-buildtool/blob/main/examples/read-sources/subprojects/netstring-c.wrap) in the example projects, using meson's subproject wrap support. Arguably that's a slightly more realistic example case than this little squaring library.

mattip:
There is a clang-tsan failure

ngoldbaum:
@mattip see #242 and #243, the failure is unrelated.

testing/cffi1/buildtool_example2/src/csrc/square.c
ngoldbaum: Thanks, I saw that on my first pass on this PR and intentionally decided to make this a simple example that didn't rely on any meson features. I think I'll make these symlinks so the two examples are identical other than how they're set up in the build.

doc/source/cffi-gen-src.rst
```diff
@@ -0,0 +1,351 @@
...
+
+   $ gen-cffi-src read-sources squared._squared squared.cdef.txt squared.csrc.c _squared.c
+
+The first positional argument is the fully qualified module name that
+will be embedded in the generated source (equivalent to the first
+argument to :meth:`FFI.set_source`).
+
+
+A Worked Example Using ``meson-python``
+=======================================
```
ngoldbaum: I think I'll leave it as-is for now.

ngoldbaum:
Sorry for taking so long to come back to this. I think this is ready for another round of review now.

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
+.. _buildtool_docs:
+
+=========================================
+Building and Distributing CFFI Extensions
+=========================================
+
+.. contents::
+
+CFFI ships a small subpackage, :mod:`cffi.buildtool`, together with a
```
mattip:
```suggestion
CFFI ships a subpackage, :mod:`cffi.buildtool`, together with a
```

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+command-line program, ``gen-cffi-src``. Both produce the same output as
```
mattip: Is the command line program installed with an entry point?

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+:meth:`FFI.emit_c_code`: a ``.c`` source file ready to be compiled into
+a CPython extension module. What they add is two convenient front-ends
+-- one that executes an existing "build" Python script, and one that
+reads a ``cdef`` and C prelude from two files. This tool enables
```
mattip: I don't understand. Can you rework the "They add" with more words?
```suggestion
a CPython extension module. They add is two convenient front-ends
-- one that executes an existing "build" Python script, and one that
reads a ``cdef`` and C prelude from two files. This tool enables
```

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+integrating with any build backend, such as `meson-python
+<https://meson-python.readthedocs.io/>`_, `scikit-build-core
+<https://scikit-build-core.readthedocs.io/>`_, or similar.
+
+The rest of this page uses meson-python in the examples, but any PEP
+517 backend that lets you run a helper program during the build can
+drive ``gen-cffi-src`` the same way.
+
+The ``cffi.buildtool`` subpackage was integrated from the
```
mattip:
```suggestion
The ``cffi.buildtool`` subpackage was vendored with permission from the
```

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Python API for ``cffi.buildtool``
+=================================
+
+.. py:module:: cffi.buildtool
+
+.. py:function:: find_ffi_in_python_script(pysrc, filename, ffivar)
+
+   Execute a Python build script and return the :class:`cffi.FFI`
+   object it defines. ``pysrc`` is the text of the script,
```
mattip: Does the build script always compile a c-extension module? Is the lib object also returned somehow?

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+
+The ``gen-cffi-src`` Command-line Tool
+======================================
+
+``gen-cffi-src`` has two subcommands. In both, the final positional
+argument is the path to the ``.c`` file to generate.
+
+.. note::
+
+   When you drive the build from a build backend, the
```
mattip: What does "drive the build" mean in this context?

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+   them through the ``dependencies:`` / ``include_directories:``
+   arguments of ``py.extension_module()``.
+
+
+``gen-cffi-src exec-python``
+----------------------------
+
+This mode takes the Python build script you would normally run by
+hand -- the one the CFFI docs show under "Main mode of usage" -- and
+generates the ``.c`` source for you. For example, given this
```
mattip: Does it stop after generating the C source, or does it also compile the c-extension module?

doc/source/cdef.rst
```diff
@@ -1,9 +1,15 @@
...
+Preparing Wrapper Modules
+=========================
 
 .. contents::
 
+.. note::
+
+   This covers how to create wrapper modules. See :ref:`buildtool_docs`
+   for instructions on how to integrate with a Python build backend and
+   distribute wrapper modules.
```
mattip: What is a wrapper module? A definition would be nice.

src/cffi/buildtool/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+This subpackage exposes a small API and a command-line entry point
+(``gen-cffi-src``) that build backends can invoke during a build to
+produce the ``.c`` source file for a CFFI extension module.:
+
+* :func:`find_ffi_in_python_script` -- execute an "exec-python" build
+  script and return the :class:`cffi.FFI` object it defines.
+* :func:`make_ffi_from_sources` -- construct an :class:`cffi.FFI`
+  from a ``cdef`` string and a C source prelude.
+* :func:`generate_c_source` -- emit the generated C source for an
+  :class:`cffi.FFI` as a string.
```
mattip: In the documentation it mentioned two modes for the subpackage: (1) generate the c code and (2) compile the c-extension and import the module. Is that mentioned somewhere here?

mattip (commented):
This seems very helpful, and will provide a path to move away from distutils/setuptools only cffi support.

Some of my comments might miss the mark. In general I think the concept of "build" is not clear enough. A cffi out-of-line module has three parts: generate C code, compile the C code into a c-extension, and import the module. Does "build" address all three? 

An in-line mode module (no compiler used) is different, I assume this does not address the inline-mode usage.

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+argument is the path to the ``.c`` file to generate.
+
+.. note::
+
+   When you drive the build from a build backend, the
+   ``libraries=``, ``library_dirs=``, ``include_dirs=``,
+   ``extra_compile_args=`` etc. arguments you pass to
+   :meth:`FFI.set_source` are *ignored*. Link and include settings are
+   the build backend's responsibility; for meson-python you express
+   them through the ``dependencies:`` / ``include_directories:``
```
rgommers: `extra_compile_args` maps to `c_args`, might want to add that for completeness

doc/source/buildtool.rst
```diff
@@ -0,0 +1,351 @@
...
+      output: '_squared.c',
+      input: ['src/squared/_squared_build.py'],
+  )
+
+  py.extension_module(
+      '_squared',
+      squared_ext_src,
+      subdir: 'squared',
+      install: true,
+      dependencies: [square_dep, py.dependency()],
```
rgommers: Drop `py.dependency` here, it's never needed inside a `py.extension_module` call. It _may_ be needed inside a `static_library` call (as above), if that used CPython C API rather than it being a pure C library (EDIT: but if it were, you'd see it in CI as a build failure).

doc/source/cdef.rst
```diff
@@ -935,10 +855,10 @@ steps.
 
 and *if* the "stuff" part is big enough that import time is a concern,
 then rewrite it as described in `the out-of-line but still ABI mode`__
-above.  Optionally, see also the `setuptools integration`__ paragraph.
+above.  Optionally, see also the :ref:`build backend and distrubution
```
rgommers: typo `distrubution` (also higher up, search the diff for it)

src/cffi/buildtool/_gen.py
```diff
@@ -0,0 +1,58 @@
+# Integrated from the cffi-buildtool project by Rose Davidson
+# (https://github.com/inklesspen/cffi-buildtool), MIT-licensed.
```
rgommers: Technically this is incorrect, it misses the required attribution. Easiest way to do this correctly is to add the upstream LICENSE file as `buildtool/LICENSE` in this PR. The main cffi license file already says _"Except when otherwise stated (look for LICENSE files in directories or
information at the beginning of each file) ..."_, so nothing else is needed.

rgommers:
I had a read through out of interest; overall this LGTM. The CLI taking the two main build modes and source as input, generated C code as output is useful and matches how Cython, f2py and other such codegen/binding tools are integrated into meson-python and scikit-build-core; the pattern should indeed work for other build backends that are able to build extension modules from C code as well.

doc/source/buildtool.rst
inklesspen: Yes; see the change in pyproject.toml.

doc/source/buildtool.rst
inklesspen: It stops after generating the C source; it is the job of a c compiler to compile the c extension module. (Build tools such as meson already have support for this.)

src/cffi/buildtool/_gen.py
inklesspen: I licensed this code under the MIT license because that was the license used for CFFI at the time that I wrote this tool. I will be happy to relicense it under the current "MIT No Attribution" license that is now used by CFFI, if that makes matters at all simpler.

doc/source/buildtool.rst
ngoldbaum: No, the build script doesn't build an extension module at all. It generates the FFI binding object (a `cffi.api.FFI` instance) which then is used to generate the C extension sources.

I guess calling it a Python build script is confusing. How about Calling it "a script that dynamically defines a set of FFI bindings"?

I'll try to rewrite these docs to make that point clearer.

doc/source/buildtool.rst
ngoldbaum: I mean when CFFI is used without CFFI's built-in distutils integration. Specifically via functionality `recompiler.py`. I'll rephrase this to make that clearer.

doc/source/buildtool.rst
ngoldbaum: I'll try to make this whole set of documentation clearer that there are no compilation steps involved - this is all about generating the C source code for a CFFI extension at build time, without relying on setuptools or CFFI's recompiler and setuptools integration.

doc/source/buildtool.rst
ngoldbaum: I'm applying your suggestion but I'm going to push a subsequent comment that rewords this.

src/cffi/buildtool/_gen.py
ngoldbaum: No worries, I think I'll just add the license file you have now to the `buildtool` subdirectory as Ralf suggested. That seems cleanest to me from an attribution point of view.

src/cffi/buildtool/__init__.py
ngoldbaum: Sorry if the language about build scripts was confusing. There's no compilation happening. The two modes of operation are either based on a Python script that defines the FFI bindings or a pair of text files that define the FFI bindings for the C API to be wrapped.

The second mode is based on the `read_sources` function that's currently not exposed in the top-level API here. I guess I might as well expose that as well if we're going to expose a Python API here.

I think we could also probably choose not to expose a public Python API here. On reflection a couple months after doing this PR initially, I lean towards not exposing a Python API and only documenting the command-line interface.

doc/source/buildtool.rst
ngoldbaum: On further reflection, I think it simplifies things to make the only UI be `python -m cffi.buildtool`, so I've removed the `gen-cffi-src` script and entry point in the latest version of the PR.

doc/source/buildtool.rst
ngoldbaum: This paragraph ended up getting dropped completely. I think the point that `cffi.buildtool` is only for generating C sources is a clearer now.

doc/source/cdef.rst
ngoldbaum: Added.

src/cffi/buildtool/_gen.py
ngoldbaum: In the latest version of the PR the whole implementation is in one `_buildtool.py` file which now includes the original MIT license verbatim in the source header.

ngoldbaum:
I think I've responded to all the review comments. I'd appreciate another look :)

doc/source/buildtool.rst
rgommers: I think the previous version you had was better. Specifically, the old version worked much better for cross compilation, because the shebang line in the installed entrypoint will correctly find the native Python interpreter that `cffi` is installed for, rather than the cross Python that's being targeted by the build. (analogous example: we use `cython` as a CLI tool)

I'd revert this particular change.

ngoldbaum: OK, fair enough, let's re-add the entry point script.

ngoldbaum: Re-added, see latest commit.

doc/source/buildtool.rst
```diff
@@ -0,0 +1,341 @@
...
+
+=========================================
+Building and Distributing CFFI Extensions
+=========================================
+
+.. contents::
+
+CFFI ships a command-line tool, ``gen-cffi-src``, that produces the
+same output as :meth:`FFI.emit_c_code`: a ``.c`` source file ready to
+be compiled into a CPython extension module. This tool enables
```
mattip:
```suggestion
be compiled into a CPython c-extension module. This tool enables
```

doc/source/buildtool.rst
```diff
@@ -0,0 +1,341 @@
...
+integrating with any build backend, such as `meson-python
+<https://meson-python.readthedocs.io/>`_, `scikit-build-core
+<https://scikit-build-core.readthedocs.io/>`_, or similar.
+
+Installing CFFI installs the ``gen-cffi-src`` script; running ``python
+-m cffi.buildtool`` invokes the same command line and behaves
```
mattip: I still find the name buildtool confusing. What does the module actually do? It does not replace mason/setuptools, does it do more than generate c code?

doc/source/buildtool.rst
```diff
@@ -0,0 +1,341 @@
...
+most useful if you already have a Python script that sets up an FFI
+definition. The second, ``read-sources`` is most useful if you are wrapping
+a large API surface and want a more structured way to specify a set of FFI
+definitions.
+
+``gen-cffi-src exec-python``
+----------------------------
+
+This mode takes a Python script that dynamically defines an FFI interface and
+accompanying C extension source code. The FFI definition script is the same
```
mattip: And does what exactly? What is the output?

doc/source/buildtool.rst
```diff
@@ -0,0 +1,341 @@
...
+
+   int square(int n);
+
+and ``squared.csrc.c``:
+
+.. code-block:: C
+
+   #include "square.h"
+
+you would run the following command to generate a CFFI extension:
```
mattip:
```suggestion
you would run the following command to generate the c code for a CFFI extension:
```

doc/source/index.rst
```diff
@@ -16,4 +16,5 @@ copy-paste from header files or documentation.
    using
    ref
    cdef
+   buildtool
```
mattip: Can you change the references to not use the word “build”?
```suggestion
   gen-cffi-src
```

src/cffi/_buildtool.py
```diff
@@ -0,0 +1,197 @@
...
+# CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+# TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+"""Implementation of the ``gen-cffi-src`` command-line tool.
+
+This module is private; the command line is the only supported
+interface. Two subcommands:
+
+``exec-python``
+    Execute a Python build script that constructs a :class:`cffi.FFI`
```
mattip:
```suggestion
    Execute a Python script that constructs a :class:`cffi.FFI`
```

src/cffi/_buildtool.py
```diff
@@ -0,0 +1,197 @@
...
+    (the same kind of script that the CFFI docs' "Main mode of usage"
+    describes) and emit the generated C source.
+
+``read-sources``
+    Build the :class:`cffi.FFI` from a separate ``cdef`` file and C
```
mattip:
```suggestion
    Create the :class:`cffi.FFI` from a separate ``cdef`` file and C
```

mattip (commented):
Sorry to be nit-picking but cffi is confusing anyway, and I think it is made more so by the use of generic “build” when a more specific concept is intended. As I understand it the intention of the tool is to expose the c-code generation mechanism provided by cffi. Then the c code is handed over to a Python compilation framework like meson or setuptools, and the resulting c-extension module can be packaged into a wheel by a wheel producing framework. I would be happier if the tool was called gen-cffi-code or gen-c-code and not buildtool. Additionally, it would be nice to scan for “build” and use a better word instead where possible.

Does that make sense or is it too restrictive?

rgommers:
> I would be happier if the tool was called gen-cffi-code or gen-c-code and not buildtool.

"buildtool" is the filename, the executable is now called `gen-cffi-src`. While that's a reasonable name: if that's being optimized, then I'd suggest starting with `cffi` (easier to identify). So `cffi-gen-src` or `cffi-gen-sources/code`.

ngoldbaum:
I just pushed a commit that globally renames `cffi.buildtool` to `cffi.gen_src` and the CLI script from `gen-cffi-src` to `cffi-gen-src`. I also tried to remove references to builds and build scripts, replacing them with references to compilation steps that happen after `cffi-gen-src` finish and FFI definition scripts. I also tried to explain better that the FFI definition script should define an FFI object named `ffibuilder` by default and how to override that default.

mattip:
Perfect. Thanks!

inklesspen:
As I promised, I've marked cffi-buildtool ([pypi](https://pypi.org/project/cffi-buildtool/1.2.0/), [github](https://github.com/inklesspen/cffi-buildtool/tree/v1.2.0)) as deprecated

PR #243, opened 2026-04-21
Update thread sanitizer CI to use a Python 3.15 image

ngoldbaum:
I can't reproduce the TSan failures under 3.15. There have been a number of thread safety fixes to the standard library since 3.14, so in addition to fixing the issue we're hitting in CI, we can also completely remove the suppressions file as well.

Also adds test timeouts per the advice [here](https://py-free-threading.github.io/testing/#configure-timeouts-to-avoid-hanging-tests).

Ping @kumaraditya303 since he originally set all this up last year.

Fixes #242.

PR #244, opened 2026-04-21
Add suppression for race seen in CI

### LINKED ISSUE #242: TSan CI is flaky
I've noticed that the TSan CI is failing regularly. See for example, this CI run on a recent PR:

https://github.com/python-cffi/cffi/actions/runs/24686671067/job/72197736127?pr=241

Unfortunately there's an issue with the docker images we're using in our tests that is preventing the tracebacks in the race report from being useful:

https://github.com/nascheme/cpython_sanity/issues/12

I'm going to try to reproduce this manually without the docker image to see if I can get more detail.

EDIT: full race report with tracebacks here: https://gist.github.com/ngoldbaum/566cc486a9e16ebb36a05c435f7944cd

ngoldbaum:
Alternative to #243 with a more minimal change.

I think simply suppressing this warning is fine, this is an upstream issue and the race doesn't seem to have any impact on CFFI itself. See https://github.com/python/cpython/issues/113956.

Closes #242.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -542,8 +542,12 @@ jobs:
 
     - name: build and install
       run: |
+        # these two steps can be removed when
+        # https://github.com/nascheme/cpython_sanity/issues/12 and the images
+        # are rebuilt
         apt update
         apt install -y llvm
+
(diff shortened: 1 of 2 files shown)


mattip:
Thanks @ngoldbaum

PR #251, opened 2026-06-29
Fix flaky pytest-run-parallel CI crash

ngoldbaum:
While triggering a full CI run to test a v2.1 release, I happened to trigger a CI crash:

https://github.com/ngoldbaum/cffi-ft/actions/runs/28403191570/job/84159323543

I tracked it down to this test, which upon inspection is clearly not thread-safe. It mutates the process global libc stdout pointer. The correct fix is to mark it as thread-unsafe.


mattip:
Thanks @ngoldbaum

PR #257, opened 2026-07-14
Fix C-API exception handling issues (lock OOM, signed char -1, exception preservation)

dynapx:
**Title:** Fix C-API exception handling issues (lock OOM, signed char -1)

---

**Description**

Hi maintainers,

This PR is a follow-up to **#256** (which fixed unchecked `PyUnicode_AsUTF8` returns). It addresses additional C-API exception handling issues identified during the same robustness audit of the `_cffi_backend` C extension.

---

### 1. `PyThread_allocate_lock` failure does not set exception

**Location:** `ffi_obj.c:ffi_init_once`

`PyThread_allocate_lock()` returns `NULL` on failure but **does not set** a Python exception. The code currently returns `NULL` without setting an exception, leading to `SystemError`.

**Fix:** Add `PyErr_SetString()` with a descriptive error message before returning.

```diff
-        if (lock == NULL)
+        if (lock == NULL)
+        {
+            PyErr_SetString(PyExc_RuntimeError,
+                    "failed to allocate thread lock for init_once");
             return NULL;
+        }
```

---

### 2. `ctypedescr_new` OOM does not set exception

**Location:** `_cffi_backend.c:new_struct_or_union_type`

`ctypedescr_new()` returns `NULL` on OOM but does not set an exception. The caller returns `NULL` without setting an exception.

**Fix:** Add `PyErr_NoMemory()`.

```diff
-    if (td == NULL)
+    if (td == NULL) {
+        PyErr_NoMemory();
         return NULL;
+    }
```

---

### 3. Signed char `-1` false negative detection

**Location:** `_cffi_backend.c:convert_from_object`

`_convert_to_char(init)` returns `-1` on error, but `-1` is also a **valid signed char value** (representing `0xFF`). The current code treats both as error, causing valid `\xFF` input to fail without setting exception.

**Fix:** Check `PyErr_Occurred()` to distinguish between error and valid `-1`.

```diff
-            if (res < 0)
+            if (res < 0 && PyErr_Occurred())
                 return -1;
```

---

### Impact

These issues can cause:
- `SystemError` when `PyThread_allocate_lock` fails
- `SystemError` when `ctypedescr_new` hits OOM
- Rejection of valid `\xFF` byte values

### DIFF
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -1762,7 +1762,7 @@ convert_from_object(char *data, CTypeDescrObject *ct, PyObject *init)
         switch (ct->ct_size) {
         case sizeof(char): {
             int res = _convert_to_char(init);
-            if (res < 0)
+            if (res < 0 && PyErr_Occurred())
                 return -1;
             data[0] = res;
             return 0;
@@ -5009,6 +5009,7 @@ static PyObject *new_struct_or_union_type(const char *name, int flag)
     int namelen = strlen(name);
     CTypeDescrObject *td = ctypedescr_new(namelen + 1);
     if (td == NULL)
+        PyErr_NoMemory();  /* ctypedescr_new does not set an exception on OOM */
         return NULL;
 
     td->ct_size = -1;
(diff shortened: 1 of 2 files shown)


mattip:
Please move to draft status until the issue is discussed.

PR #265, opened 2026-07-23
Fix variadic calls on Apple platforms with a non-OS libffi

### LINKED ISSUE #264: Variadic calls on arm64 Apple builds with a non-SDK libffi are broken
Any `_cffi_backend` built on an Apple platform against a libffi **other than the one shipped with the macOS SDK** never calls `ffi_prep_cif_var()`. Variadic functions are then called through a cif prepared by `ffi_prep_cif()`, i.e. with the fixed-arguments calling convention. On arm64 Apple platforms variadic arguments are passed on the stack, unlike named arguments, so the callee reads garbage: integer varargs arrive as random values, and pointer varargs typically crash the process.

This affects:

* all iOS wheels published on PyPI (cffi 2.0.0 and 2.1.0; both simulator and device) — they are built against the static libffi from `beeware/cpython-apple-source-deps`, per `.github/workflows/ci.yaml`;
* macOS builds where a Homebrew libffi's headers end up ahead of the SDK's (`setup.py` actively arranges `PKG_CONFIG_PATH` for this in `use_homebrew_for_libffi()`).

Not affected: macOS wheels from PyPI (built against the SDK's libffi headers), and `ctypes` (CPython calls `ffi_prep_cif_var()` unconditionally when extra args are passed — which is also a handy control experiment on the same device).

## Reproducer

On an iOS simulator (e.g. via CPython's iOS testbed) with a PyPI cffi wheel:

```python
from cffi import FFI

ffi = FFI()
ffi.cdef("int snprintf(char *str, size_t size, const char *format, ...);")
libc = ffi.dlopen(None)

buf = ffi.new("char[256]")
s = ffi.new("char[]", b"hello")
rc = libc.snprintf(buf, 256, b"[%s] [%d]", s, ffi.cast("int", 42))
print(rc, ffi.string(buf))   # expected: 12 b'[hello] [42]'
```

Actual result: the process crashes with `EXC_BAD_ACCESS` — `snprintf` receives a garbage pointer for `%s`:

```
_platform_strlen +4        <- strlen(garbage)
__vfprintf
_vsnprintf
snprintf +100
_cffi_backend  ffi_call_SYSV +80
_cffi_backend  ffi_call_int +1580
_cffi_backend  ffi_call +52
_cffi_backend  cdata_call +2628
```

The same call via `ctypes` (declaring only the three fixed parameters in `argtypes`) works correctly in the same process.

## Root cause

`src/c/_cffi_backend.c`:

```c
#elif defined(__APPLE__) && defined(FFI_AVAILABLE_APPLE)

# define CFFI_CHECK_FFI_PREP_CIF_VAR __builtin_available(macos 10.15, ios 13, watchos 6, tvos 13, *)
# define CFFI_CHECK_FFI_PREP_CIF_VAR_MAYBE 1

...
#else

# define CFFI_CHECK_FFI_PREP_CIF_VAR 0
# define CFFI_CHECK_FFI_PREP_CIF_VAR_MAYBE 0

#endif
```

`FFI_AVAILABLE_APPLE` is defined only by the libffi header shipped with the macOS SDK (it carries Apple's availability annotations). Building against any other libffi — the static one used for the iOS wheels, or Homebrew's — leaves it undefined, so the build falls into the `#else` branch and compiles out the `ffi_prep_cif_var()` path entirely (along with `ffi_closure_alloc()` / `ffi_prep_closure_loc()`). The bundled libffi *does* provide all of these functions; they are simply never used.

Binary-level confirmation on the PyPI 2.1.0 simulator wheel — the symbol is present but has no call sites, while CPython's `_ctypes` (which works) calls it:

```console
$ otool -tv _cffi_backend.cpython-313-iphonesimulator.so | grep -c 'bl.*_ffi_prep_cif_var'
0
$ otool -tv _ctypes.cpython-313-iphonesimulator.so | grep -c 'bl.*_ffi_prep_cif_var'
1
```

## Fun detail: the cffi test suite already caught this

`src/c/test_c.py` contains two iOS markers for "unknown reasons" that are both
this bug:

* `test_call_function_9` — `xfail(is_ios, reason="For an unknown reason f(1, cast(BInt, 42)) returns 36792864")`: an integer vararg read from a garbage location;
* `test_FILE` — `skipif(is_ios, reason="For an unknown reason fscanf() doesn't read anything on 3.14 and crashes on 3.13")`: `fscanf()` takes pointer varargs, hence the crash.

ambv:
Closes #264

The CFFI_CHECK_FFI_* feature checks only enabled ffi_prep_cif_var(), ffi_closure_alloc() and ffi_prep_closure_loc() on Apple platforms when FFI_AVAILABLE_APPLE was defined.  That macro comes from the libffi header shipped with the macOS SDK, so any build against a different libffi - the static libffi that iOS/tvOS/watchOS wheels are built against, or a Homebrew libffi on macOS - silently fell into the fallback branch that disables all three functions.

Without ffi_prep_cif_var(), variadic functions are called through a cif prepared by ffi_prep_cif(), i.e. with the fixed-arguments calling convention.  On arm64 Apple platforms variadic arguments are passed on the stack, unlike named arguments, so the callee read garbage: integers came back as random values and pointer arguments typically crashed (e.g. any printf-family callee dereferencing a garbage %s).

There is no runtime-availability question in this configuration: the libffi the module was linked against always provides all three functions, so the new branch enables them unconditionally.

The two "for an unknown reason" iOS markers in test_c.py were this bug: test_call_function_9 (garbage int return) and test_FILE (fscanf() reading nothing on 3.14, crashing on 3.13).  Remove the markers so both act as regression tests.  Verified on an arm64 iPhone simulator with CPython 3.13: the full test_c.py passes (208 passed, 21 skipped), and both tests fail again when the fix is reverted.

### DIFF
--- a/doc/source/whatsnew.rst
+++ b/doc/source/whatsnew.rst
@@ -8,10 +8,9 @@ v2.2.0.dev0
 * Fixed variadic function calls on Apple platforms when ``_cffi_backend`` is
   built against a libffi other than the one provided by the macOS SDK — that
   is, the iOS/tvOS/watchOS wheels, or a macOS build using a Homebrew libffi.
-  Such builds never used ``ffi_prep_cif_var()``, so variadic arguments were
-  passed using the fixed-arguments calling convention; on arm64 they are
-  passed differently, so the callee read garbage: integer arguments arrived
-  as random values, and pointer arguments typically caused a crash.
+  (`#265`_).
+
+.. _`#265`: https://github.com/python-cffi/cffi/pull/265
 
 v2.1.0
 ======
--- a/doc/source/whatsnew.rst
+++ b/doc/source/whatsnew.rst
@@ -5,7 +5,13 @@ What's New
 v2.2.0.dev0
 ===========
 
-* TBD
+* Fixed variadic function calls on Apple platforms when ``_cffi_backend`` is
+  built against a libffi other than the one provided by the macOS SDK — that
+  is, the iOS/tvOS/watchOS wheels, or a macOS build using a Homebrew libffi.
+  Such builds never used ``ffi_prep_cif_var()``, so variadic arguments were
+  passed using the fixed-arguments calling convention; on arm64 they are
+  passed differently, so the callee read garbage: integer arguments arrived
+  as random values, and pointer arguments typically caused a crash.
 
 v2.1.0
 ======
--- a/src/c/_cffi_backend.c
+++ b/src/c/_cffi_backend.c
@@ -117,6 +117,25 @@
 # define CFFI_CHECK_FFI_PREP_CIF_VAR __builtin_available(macos 10.15, ios 13, watchos 6, tvos 13, *)
 # define CFFI_CHECK_FFI_PREP_CIF_VAR_MAYBE 1
 
+#elif defined(__APPLE__)
+
+/* On Apple platforms, but built against a libffi whose headers do not
+ * define FFI_AVAILABLE_APPLE: not the OS-provided libffi, but e.g. the
+ * static libffi that iOS/tvOS/watchOS wheels are built against, or a
+ * Homebrew libffi on macOS.  There is no runtime-availability question
+ * there: whatever the linker resolved is present, so all three functions
+ * can be used unconditionally.  ffi_prep_cif_var() in particular must be
+ * used: on arm64 Apple platforms variadic arguments are passed on the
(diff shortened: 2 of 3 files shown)


mattip:
Thanks @ambv. This could enable some of the skipped tests in test_c.py, but not needed in this PR.

PR #268, opened 2026-07-28
Fix Generated C depends on the host interpreter

### LINKED ISSUE #267: `cffi-gen-src` output varies depending on running interpreter
While working on https://github.com/hynek/argon2-cffi-bindings/pull/137, I noticed that the output of `cffi-gen-src` is different for a free-threaded and GIL-enabled interpreter.

Self-contained reproducer:

```
$ cat demo_build.py
from cffi import FFI
ffibuilder = FFI()
ffibuilder.cdef("int square(int n);")
ffibuilder.set_source("_demo", "static int square(int n) { return n * n; }")
$
$ python3.14 -m cffi.gen_src exec-python demo_build.py demo-gil.c
$ python3.14t -m cffi.gen_src exec-python demo_build.py demo-ft.c
$ diff -u demo-gil.c demo-ft.c
--- demo-gil.c	2026-07-28 14:57:58
+++ demo-ft.c	2026-07-28 14:58:20
@@ -1,3 +1,4 @@
+#define _CFFI_NO_LIMITED_API
 #define _CFFI_

 /* We try to define Py_LIMITED_API before including Python.h.
```

ngoldbaum:
Fixes #267.

Removed code to disable limited API builds on free-threaded Python 3.13 and 3.14, which makes the generated source code from `ffi.emit_c_code()` depend on the emitting interpreter.

After doing that, I noticed that the `USE_LIMITED_API` constant in `recompiler.py` is vestigial (to fix issues on Python 3.4 and older on Windows) and can be deleted along with this change.


testing/cffi1/test_cffi_gen_src.py
```diff
@@ -211,3 +211,16 @@ def test_import_is_inert():
     assert proc.returncode == 0, proc.stderr
     assert proc.stdout == ""
     assert proc.stderr == ""
+
+
+def test_output_independent_of_generating_interpreter(tmp_path):
```
mattip: Using the `run_cffi_gen_src` and calling (intead of _run) `run_cffi_gen_src("exec-python", ...` like elsewhere in the file might be nice

mattip (approved):
Thanks, nice. 

LGTM, one minor nit. Maybe not worth changing.

testing/cffi1/test_cffi_gen_src.py
ngoldbaum: Thanks for the suggestion, I agree that reads better. I also trimmed the comment a bit.

mattip:
Thanks @ngoldbaum

Issue #263, opened 2026-07-21
Binary (windows?) wheels for 3.15 are probably incompatible with newly-released 3.15b4; ~perhaps a 3.15b4 incompatibility in general~


jamadden:
Over in [this cython bug](https://github.com/cython/cython/issues/7847) I documented a case where binary wheels that worked in 3.15b3 failed in 3.15b4, which came out this weekend. 

I can't prove it, but I'm pretty sure CFFI's wheels are in the same situation. I believe this because, without any related code changes on gevent's part, [our Windows builds that worked correctly on 3.15b3](https://github.com/gevent/gevent/actions/runs/28617489479/job/84864815365) suddenly started [segfaulting in nearly every test on 3.15b4](https://github.com/gevent/gevent/actions/runs/29860621690). 

~There is some additional information. Even, after I started building CFFI from scratch on mac/linux (`pip install --no-binary=:all: cffi`),~ (EDIT: I think I must still have been using the pre-built wheel) I was seeing one crash on 3.15 that [_appears_ related to CFFI](https://github.com/gevent/gevent/actions/runs/29858372114/job/88728378510#step:18:234). That's the only test I had that uses CFFI (the Windows tests _all_ use CFFI), and when I disabled that test, the crashes went away, so I think we can rule out gevent or 3.15b4 itself.

mattip:
Are you set up to run your tests under a debugger on those platforms? Typically that kind of issue crops up when an application's cffi-allocated variable is not held in the interpreter and the GC collects the buffer while something else is accessing that buffer. It would help if you could cut the failure down to a small reproducer.

jamadden:
I can reproduce one of the smaller crashes I saw locally on macOS (I have no access to Windows). Using the pre-built PyPI wheel on 3.15b4, ``lldb python -m gevent.tests.test__core`` crashes like so:

```
* thread #2, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
    frame #0: 0x00000001013709b4 Python`take_gil + 60
Python`take_gil:
->  0x1013709b4 <+60>: ldr    x22, [x20, #0x10]
    0x1013709b8 <+64>: add    x0, x22, #0x50
    0x1013709bc <+68>: bl     0x1014c4da4    ; symbol stub for: pthread_mutex_lock
    0x1013709c0 <+72>: cbnz   w0, 0x101370c6c ; <+756>
Target 0: (Python) stopped.
(lldb) bt
[...]

PR #269, opened 2026-07-30
Use PyGILState_Ensure to avoid accessing CPython internals

ngoldbaum:
Following a suggestion from @kumaraditya303, this removes direct access to the `PyThreadState` struct in `gil_ensure()` and replaces it with `PyGILState_Ensure`.

The direct struct access is the cause of the crash in #263.

The current code dates back to Python 3.6 and works around differences spanning the Python 2->3 transition. That is all historical now. The code that is currently active is identical to the content of `PyGILState_Ensure`.

### DIFF
--- a/src/c/call_python.c
+++ b/src/c/call_python.c
@@ -1,6 +1,6 @@
 static PyObject *_current_interp_key(void)
 {
-    PyInterpreterState *interp = PyThreadState_GET()->interp;
+    PyInterpreterState *interp = PyInterpreterState_Get();
     return PyInterpreterState_GetDict(interp);   /* shared reference */
 }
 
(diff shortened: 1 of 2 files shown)


src/c/call_python.c
```diff
@@ -17,13 +17,17 @@ static PyObject *_get_interpstate_dict(void)
     int err;
     PyInterpreterState *interp;
 
-    tstate = PyThreadState_GET();
+#if PY_VERSION_HEX >= 0x030D0000
+    tstate = PyThreadState_GetUnchecked();
+#else
+    tstate = _PyThreadState_UncheckedGet();
```
nitzmahone, line 23: I haven't had time to think through if this an issue (or even possible), but looking at https://github.com/python/cpython/issues/131346 - if this were called from <3.12 without the GIL held, IIUC it would return a stale/disconnected thread state instead of bailing out.

ngoldbaum: `_get_interpstate_dict()` has exactly two callers, both of which hold the GIL: 

* `_ffi_def_extern_decorator` (call_python.c:115) — runs as an ordinary Python builtin call (@ffi.def_extern());
* `_update_cache_to_call_python` (call_python.c:158) — called only from cffi_call_python at line 264, which is after gil_ensure() at line 255.

So it's not a problem in the current codebase. Think it deserves a comment?

nitzmahone: Nah, that's the part I hadn't worked out yet, so seems like it'll be fine- IIRC the original version of the code would've fatal-errored if you somehow managed to get into that situation anyway (so Don't Do That 😆 )

kumaraditya303 (approved):
LGTM

Issue #229, opened 2026-02-13
ValueError: too many values to unpack


qris:
```
pdm.exceptions.BuildError: Build backend raised error: Showing the last 10 lines of the build output:
    ffibuilder.compile(verbose=True)
  File "/tmp/pdm-build-env-ot45gqlp-overlay/lib/python3.10/site-packages/cffi/api.py", line 727, in compile
    return recompile(self, module_name, source, tmpdir=tmpdir,
  File "/tmp/pdm-build-env-ot45gqlp-overlay/lib/python3.10/site-packages/cffi/recompiler.py", line 1581, in recompile
    outputfilename = ffiplatform.compile('.', ext,
  File "/tmp/pdm-build-env-ot45gqlp-overlay/lib/python3.10/site-packages/cffi/ffiplatform.py", line 20, in compile
    outputfilename = _build(tmpdir, ext, compiler_verbose, debug)
  File "/tmp/pdm-build-env-ot45gqlp-overlay/lib/python3.10/site-packages/cffi/ffiplatform.py", line 50, in _build
    [soname] = cmd_obj.get_outputs()
ValueError: too many values to unpack (expected 1)
```

I know there's probably something wrong with my build, but this isn't supposed to happen, and probably makes it harder to debug.

This is CFFI 2.0.0.

mattip:
The first step toward solving this would be a clear reproducer, hopefully not too large

PR #273, opened 2026-08-30
Use get_ext_fullpath instead of get_outputs in ffiplatform._build

### LINKED ISSUE #246: "too many values to unpack" when there are extension modules in pyproject.toml
When calling `ffi.verify("#include <string.h>")` in a Python project, cffi threw the following exception:

```python
  /workspace/cle/tests/test_clemory.py(22)test_cclemory()
     20         int memcmp(const void* s1, const void* s2, size_t n);
     21     """)
---> 22     c = ffi.verify("""
     23         #include <string.h>
     24     """)
  /workspace/angr-venv/lib/python3.14/site-packages/cffi/api.py(468)verify()
    466         # Make a Verifier() and use it to load the library.
    467         self.verifier = Verifier(self, source, tmpdir, **kwargs)
--> 468         lib = self.verifier.load_library()
    469         #
    470         # Save the loaded library for keep-alive purposes, even
  /workspace/angr-venv/lib/python3.14/site-packages/cffi/verifier.py(105)load_library()
    103                     if not self._has_source:
    104                         self._write_source()
--> 105                     self._compile_module()
    106             return self._load_library()
    107
  /workspace/angr-venv/lib/python3.14/site-packages/cffi/verifier.py(201)_compile_module()
    199         # compile this C source
    200         tmpdir = os.path.dirname(self.sourcefilename)
--> 201         outputfilename = ffiplatform.compile(tmpdir, self.get_extension())
    202         try:
    203             same = ffiplatform.samefile(outputfilename, self.modulefilename)
  /workspace/angr-venv/lib/python3.14/site-packages/cffi/ffiplatform.py(20)compile()
     18     saved_environ = os.environ.copy()
     19     try:
---> 20         outputfilename = _build(tmpdir, ext, compiler_verbose, debug)
     21         outputfilename = os.path.abspath(outputfilename)
     22     finally:
> /workspace/angr-venv/lib/python3.14/site-packages/cffi/ffiplatform.py(50)_build()
     48             dist.run_command('build_ext')
     49             cmd_obj = dist.get_command_obj('build_ext')
---> 50             [soname] = cmd_obj.get_outputs()
     51         finally:
     52             set_threshold(old_level)
```

`cmd_obj.get_outputs()` returns:

```
ipdb> pp cmd_obj.get_outputs()
['/workspace/cle/tests/__pycache__/_cffi__x4038f167x5672c41f.cpython-314-x86_64-linux-gnu.so',
 '/workspace/cle/tests/__pycache__/uefi_firmware/efi_compressor.cpython-314-x86_64-linux-gnu.so']
```

Looks like this is because I have another native extension in this project, and `_build()` does not expect other extensions to exist. My local workaround is changing `[soname] = cmd_obj.get_outputs()` to `soname = cmd_obj.get_outputs()[0]`.

Please let me know if this is a workaround that you are comfortable with, and I can submit a PR if you like!

afonsojanu:
Closes #246. Related to #229, which hit the same underlying crash.

`_build()` compiles a single extension and then unpacks its output with `[soname] = cmd_obj.get_outputs()`. That works as long as `build_ext` reports exactly one output, but `get_outputs()` reflects everything the command built, not just the extension we asked for. #246 shows a project with another native extension already registered ending up with two entries there, so the unpacking raises a bare `ValueError: too many values to unpack` instead of anything actionable. #229 is a beginner hitting the identical crash with no clue what went wrong, which is basically the same request: give a real answer instead of an opaque unpack failure.

`build_ext.get_ext_fullpath(name)` is the standard way to ask for the output path of one specific extension, so swapping to that sidesteps the whole problem: we get back exactly the file for the extension we passed in, regardless of what else `get_outputs()` would have reported.

Added a regression test in `testing/cffi0/test_platform.py` that monkeypatches `build_ext.get_outputs` to return an extra, nonexistent path (simulating the multi-extension situation from #246) and checks that `ffi.verify()` still compiles and loads correctly. Confirmed it reproduces the exact reported `ValueError: too many values to unpack (expected 1)` against the unpatched `_build()`, and passes with the fix. Ran the full `testing/cffi0/test_platform.py`, `test_zdistutils.py`, `test_verify.py` and `test_verify2.py` suites locally, all green.


testing/cffi0/test_platform.py
```diff
@@ -23,3 +26,33 @@ def test_flatten():
...
+    def get_outputs_with_extra_entry(self):
+        return list(original_get_outputs(self)) + ['/nonexistent/other.so']
+
+    monkeypatch.setattr(real_build_ext, 'get_outputs',
+                        get_outputs_with_extra_entry)
+
+    # force a fresh module name/compile every run, so the monkeypatched
+    # get_outputs() above is actually exercised instead of reusing a
+    # previously-built module cached under the same checksum-derived name
+    tag = binascii.hexlify(os.urandom(8)).decode()
```
mattip: This seems simpler and does not require importing binascii. 

```suggestion
    tag = os.urandom(8).hex()
```

mattip (commented):
One small nit. LGTM, if CI passes.

testing/cffi0/test_platform.py
afonsojanu: Good call, pushed that.

testing/cffi0/test_platform.py
```diff
@@ -23,3 +25,33 @@ def test_flatten():
...
+
+@pytest.mark.thread_unsafe(reason="monkeypatches a shared distutils class method")
+def test_compile_with_extra_build_ext_outputs(monkeypatch):
+    # Some setuptools/distutils versions can make build_ext.get_outputs()
+    # return more entries than the single extension we asked it to build
+    # (see https://github.com/python-cffi/cffi/issues/246, which is the
+    # same underlying unpacking crash reported in
+    # https://github.com/python-cffi/cffi/issues/229). Unpacking that list
+    # unconditionally used to raise a confusing
+    # "ValueError: too many values to unpack".
```
mattip, line 37:
```suggestion
    # Some setuptools/distutils versions can make build_ext.get_outputs()
    # return more than one entry
```

mattip: AI can be too verbose

PR #274, opened 2026-09-04
Call PyType_Ready() on ThreadCanary_Type

emontnemery:
### Proposed change

Call PyType_Ready() on ThreadCanary_Type before the first `thread_canary` is created. Placed in `thread_canary_register()`.

### Background

ThreadCanary_Type (src/c/misc_thread_common.h) is declared with PyVarObject_HEAD_INIT(NULL, 0) and is never passed to PyType_Ready(), nor is its ob_type set explicitly - so its metatype stays NULL for the life of the process. Yet instances are created: thread_canary_register() does PyObject_New(ThreadCanaryObj, &ThreadCanary_Type) and stores the object in the foreign thread's PyThreadState dict.

Code that inspects such an instance's type dereferences the NULL metatype and crashes; in my case we get crashes when using zhuyifei1999/guppy3 to walk the heap.

#### CPython reference

The NULL in PyVarObject_HEAD_INIT(NULL, 0) is only a placeholder for compilers that reject &PyType_Type as a static initializer; it isn't meant to survive to runtime. Per the CPython C-API docs for the type ob_type field (https://docs.python.org/3/c-api/typeobj.html#pyobject-slots):

> …the convention is to pass NULL to the PyObject_HEAD_INIT macro and to initialize this field explicitly at the start of the module's initialization function, before doing anything else. […] This should be done before any instances of the type are created. PyType_Ready() checks if ob_type is NULL, and if so, initializes it to the ob_type field of the base class.


mattip:
Why are you inspecting that type? Is this some kind of test harness?

emontnemery:
We see crashes when calling the Home Assistant [`profiler.memory`](https://www.home-assistant.io/actions/profiler.memory/) action, which uses guppy3 to do a [heap walk and dump the output to a file](https://github.com/home-assistant/core/blob/e5a24a74b47db64065d21d149149fe7b4e7957bd/homeassistant/components/profiler/services.py#L468-L469). The purpose of that action is to aid debugging of memory leaks etc.

guppy3 doesn't guard against NULL metatype and segfaults (I've opened a PR on guppy3 to guard against that https://github.com/zhuyifei1999/guppy3/pull/56)

My understanding is that live CPython objects should not have a NULL metatype, or is that intentionally the case here for some reason?

mattip:
It is, as the name says, a canary probe. Maybe we should make it only live in tests, nothing at runtime uses it.

emontnemery:
> It is, as the name says, a canary probe. Maybe we should make it only live in tests, nothing at runtime uses it.

OK, that makes sense 👍 
Should I close this PR and replace it with one which hides it in runtime?

Edit: I seems to me the canary was [added ](https://foss.heptapod.net/pypy/cffi/-/commit/73a16cc62771ecb5a4b58837b44c7dde4844f26a) to fix https://foss.heptapod.net/pypy/cffi/-/work_items/362, and is not there to aid testing.

https://github.com/python-cffi/cffi/blob/1f2d6ece55319e6016b97f424d264c47b505e88a/src/c/misc_thread_common.h#L32-L50

https://github.com/python-cffi/cffi/blob/1f2d6ece55319e6016b97f424d264c47b505e88a/src/c/misc_thread_common.h#L52-L82

mattip:
OK, then the PyType_Ready should probably be called next to the type object creation, around line 214. Does that work? https://github.com/python-cffi/cffi/blob/61fe449bd5bf9ae48211e822a7eefa5cb07c11d3/src/c/misc_thread_common.h#L210-L220

emontnemery:
> OK, then the PyType_Ready should probably be called next to the type object creation, around line 214. Does that work?

Done.

mattip:
Thanks @emontnemery

emontnemery:
@mattip thanks a lot for the quick merge! Do you have plans to make a release anytime soon?

PR #276, opened 2026-09-09
riscv64: add manylinux, musllinux to matrix

threexc:
This adds riscv64 builds for both manylinux and musllinux to the CI matrix. Some Python packages run into issues building for musllinux (that wasn't the case here), so I separated the additions into distinct commits to make a rebase/revert easier if needed.

[Here](https://github.com/threexc/cffi/actions/runs/34298857623) is a dry run I did on my fork, with these changes and extra `revertme` commits to only run riscv64 jobs and to restrict the triggers.


threexc:
My mistake, @justeph beat me to it. Thanks!

PR #277, opened 2026-09-09
workflows: ci: add riscv64 musllinux, stable ABI builds

threexc:
Following up the recent merge from https://github.com/python-cffi/cffi/pull/227, add musllinux riscv64 builds for cffi.

Here is a dry run on my fork, with extra `revertme` commits added to strip out other architectures/platforms for testing: 

https://github.com/threexc/cffi/actions/runs/34350063320

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -152,6 +152,16 @@ jobs:
           - { spec: cp314-manylinux_riscv64, arch: riscv64, omit: ${{ env.skip_slow_jobs }} }
           - { spec: cp314t-manylinux_riscv64, arch: riscv64, omit: ${{ env.skip_slow_jobs }} }
 
+          # riscv64 musllinux
+          - { spec: cp310-musllinux_riscv64, arch: riscv64, omit: ${{ env.skip_ci_redundant_jobs }} }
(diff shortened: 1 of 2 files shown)


mattip:
Which platform is more popular in the niche riscv64 world, manylinux or muslinux? Do we have any PyPI download statistics that might give us an idea if this is really necessary? We already have 26 workflows for this package. Maybe we should poke again at the idea of a limited api abi3 wheel, that would make this all more palatable.

threexc:
> Which platform is more popular in the niche riscv64 world, manylinux or muslinux? Do we have any PyPI download statistics that might give us an idea if this is really necessary? We already have 26 workflows for this package. Maybe we should poke again at the idea of a limited api abi3 wheel, that would make this all more palatable.

For now it is definitely manylinux, but getting cffi wheels for musllinux would also go a long way to improving Python compatibility in that area, if our experiences adding riscv64 wheels in the RISE registry (and upstream, with Rust-based packages) are any indicator. I don't have numbers on musllinux consumption specifically. We can certainly look into the ab3 wheel.

threexc:
There are a lot of errors on my initial rework to add abi3 wheels, so I'll convert this back to a draft for now.

mattip:
Hah, yeah, it is definitely not a simple flip-a-flag change. My inclination is to leave riscv64-muslinux out of the mix for now.

PR #278, opened 2026-09-14
Uniformly disable assertions on release builds and enable them for tests

ngoldbaum:
Currently the Linux and iOS wheels are built with debug assertions enabled. This could lead to aborts at runtime for release builds if someone happens to hit an assert. The Mac and Windows wheels do not have asserts.

I'm pretty sure this was unintentional: setuptools 75.7.0 and newer no longer inherit `-DNDEBUG` from CPython.

This PR sets up an intentional policy: release wheels shouldn't have asserts but wheels built for testing and other test jobs should have asserts enabled. Also sets up CI scaffolding and a test for this.

I have a followup PR that uses asserts to document and enforce invariants which led me to notice and think about this.


mattip:
Makes sense, and the test is passing. Thanks @ngoldbaum

PR #279, opened 2026-09-21
Document header dependencies for source distributions

### LINKED ISSUE #168: DOC: document use of 'depends' kwarg to include headers in setuptools call
When compiling a CFFI module that compiles source code ([these instructions](https://cffi.readthedocs.io/en/latest/overview.html#id19)), if that source code contains headers, and those headers are used in the C code, it fails with the default invocation of `python -m build`. This is because CFFI is not able to tell `setuptools` about the header files, and by default `python -m build` generates an isolated enviroment containing only the files that `setuptools` knows about, which will not include the required headers.

A workaround is to pass something like `package_data={"": ["*.h"]},` to `setuptools.setup`, which will grab all `.h` files, but it may be worth either a) documenting this behaviour or b) augmenting `set_source` with a headers parameter so it can mark those headers as being required.

saitakarcesme:
Document `set_source(..., depends=[...])` for project-local headers and distinguish it from `include_dirs`. A header can be found when compiling in the working tree but be absent when a wheel is built from an sdist. The overview example now declares `pi.h` as a dependency and includes the project directory in the compiler search path.

Fixes #168, following the discussion's preference to document the existing setuptools keyword.

Validation: built a minimal CFFI project through sdist → wheel with the checked-out CFFI. Without `depends`, `pi.h` was absent from the tarball and compilation failed; with it, the wheel built successfully, installed, and its C function returned the expected result. Both cases used `include_dirs=['.']`. `sphinx-build -W -b html doc/source doc/build/html` and `git diff --check` pass.

Disclosure: documentation and build verification prepared using OpenAI Codex in an authorized automated contribution session.


mattip:
Thanks @saitakarcesme

